PR #525, opened 2017-03-30
Surveys

edvardvb:
Edit by Finn:
There is more functionality to be added, such as implementing template prod fixtures and perhaps creating endpoints for creating and editing individual questions and options. However, in the spirit of smaller PRs, maybe we should try to get the current functionality reviewed and merged before any of that?


finnss:
https://github.com/webkom/lego-webapp/pull/901

finnss:
Creating and answering surveys (=creating submission) should work for the most part now. I think. On the other hand, it's 5am, so who really knows 🙃

finnss:
Everything except surveys and submissions is nested, so you create entries as such:

Survey:
```
{
    "title": "Test",
    "is_clone": false,
    "event": 1,
    "questions": [
      {
        "question_text": "Spørsmål 1, Radio Button",
        "question_type": 1,
        "alternatives": [
          {
            "alternative_text": "Ja",
            "alternative_type": 1
          },
          {
            "alternative_text": "Nei",
            "alternative_type": 1
          }
        ]
      },
      {
        "question_text": "Spørsmål 2, Checkbox",
        "question_type": 2,
        "alternatives": [
          {
            "alternative_text": "Alt 1",
            "alternative_type": 2
          },
          {
            "alternative_text": "Alt 2",
            "alternative_type": 2
          }
        ]
      },
      {
        "question_text": "Spørsmål 3, Textbox",
        "question_type": 3
      }
    ]
}
``` 

Submission (tilhørende survey gitt via url):
```
{
    "user": 1,
    "answers": [
      {
        "question": 1,
        "selected_options": [1]
      },
      {
        "question": 2,
        "selected_options": [1, 2]
      },
      {
        "question": 3,
        "answer_text": "Bra arrangement!"
      }
    ]
}
```

finnss:
There is more functionality to be added, such as fixing `submit` and implementing templates and perhaps creating endpoints for creating and editing individual questions and options. However, in the spirit of smaller PRs, maybe we should try to get the current functionality reviewed and merged before any of that?

lego/apps/surveys/serializers.py
```diff
@@ -0,0 +1,144 @@
...
+    class Meta:
+        model = Survey
+        fields = ('id', 'title', 'active_from', 'event')
+
+
+class SubmissionReadSerializer(BasisModelSerializer):
+    # Do we want user and survey as nested objects?
+    # user = PublicUserSerializer()
+    # survey = SurveyReadSerializer()
+    answers = AnswerSerializer(many=True)
```
edvardvb: Do we need answers in the list view?

lego/apps/surveys/views.py
```diff
@@ -0,0 +1,49 @@
...
+        if self.action in ['create', 'update', 'partial_update']:
+            return SurveyCreateAndUpdateSerializer
+        elif self.action in ['retrieve']:
+            return SurveyReadDetailedSerializer
+        return SurveyReadSerializer
+
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+        survey = Survey.objects.get(instance.id)
+        survey.delete()
```
edvardvb: Use `is_answered_by_someone` here? Not sure if we want that or not :stuck_out_tongue: Would definitively be nice on updates to avoid changing a survey after it's active

finnss: I'm thinking it should be possible for someone who's able to delete surveys to do so whenever they want. There should be some sort of validation for sure, but that can be in the front-end, the backend shouldn't really prevent it imo.

lego/apps/surveys/serializers.py
finnss: Remember that this is the Submission list view, not the Survey list view. This is the endpoint that will be called when the user click a specific survey and navigates to "Answers". I suppose this view should actually just return some aggregated data, such as number of submissions and how many users selected each option for each question.

lego/apps/surveys/permissions.py
```diff
@@ -0,0 +1,30 @@
...
+        survey = Survey.objects.get(id=view.kwargs['survey_pk'])
+        event = getattr(survey, 'event')
+        user = request.user
+        user_attended_event = Registration.objects.filter(
+            event=event.id, user=user.id, presence=constants.PRESENT).exists()
+
+        if user.has_perm(EDIT, obj=Survey):
+            return True
+        if view.action in ['create']:
+            return user_attended_event
```
finnss: Do we want to add a check here to make sure each user can only answer once per survey?

lego/apps/users/fixtures/test_abakus_groups.yaml
```diff
@@ -92,11 +92,13 @@
 - model: users.abakusgroup
   pk: 19
   fields: {deleted: false, name: Bedkom, description: '', parent: 18, logo: null,
-    type: komite, text: '', permissions: '[]', lft: 3, rght: 4, tree_id: 15, level: 2}
+    type: komite, text: '', permissions: '["/sudo/admin/companyinterest/", "/sudo/admin/surveys/",
+    "/sudo/admin/submissions/"]', lft: 3, rght: 4, tree_id: 15, level: 2}
```
finnss: These will have to be added in prod manually, right?

lego/apps/surveys/permissions.py
finnss: Submission already has `unique_together = (survey, user)`, but perhaps checking for it here would provide a better error message?

lego/apps/surveys/serializers.py
```diff
@@ -0,0 +1,160 @@
...
+        survey = Survey.objects.get(pk=self.context['view'].kwargs['survey_pk'])
+        answers = None if 'answers' not in validated_data else validated_data.pop('answers')
+        submission = Submission.objects.create(survey=survey, **validated_data)
+        submission.save()
+
+        if answers is not None:
+            for answer in answers:
+                question = answer.pop('question')
+
+                if getattr(question, 'question_type') in [1, 2]:
```
edvardvb: Remove as discussed

lego/apps/surveys/serializers.py
```diff
@@ -0,0 +1,160 @@
...
+
+    class Meta:
+        model = Submission
+        fields = ('id', 'user', 'answers')
+
+    def create(self, validated_data):
+        survey = Survey.objects.get(pk=self.context['view'].kwargs['survey_pk'])
+        answers = None if 'answers' not in validated_data else validated_data.pop('answers')
+        submission = Submission.objects.create(survey=survey, **validated_data)
+        submission.save()
```
edvardvb: unecessary

lego/apps/surveys/serializers.py
```diff
@@ -0,0 +1,160 @@
...
+            for answer in answers:
+                question = answer.pop('question')
+
+                if getattr(question, 'question_type') in [1, 2]:
+                    if getattr(question, 'question_type') is 1 and \
+                                    len(answer['selected_options']) > 1:
+                        raise exceptions.ValidationError('You cannot select multiple options for '
+                                                         'this type of question.')
+
+                    Answer.create(submission, question, **answer)
```
edvardvb: Remove and fix as discussed (99-103)

lego/apps/surveys/serializers.py
```diff
@@ -0,0 +1,160 @@
...
+        answers = None if 'answers' not in validated_data else validated_data.pop('answers')
+        submission = Submission.objects.create(survey=survey, **validated_data)
+        submission.save()
+
+        if answers is not None:
+            for answer in answers:
+                question = answer.pop('question')
+
+                if getattr(question, 'question_type') in [1, 2]:
+                    if getattr(question, 'question_type') is 1 and \
```
edvardvb: Any reason to do this and not just `question.question_type is 1...`?

lego/apps/surveys/serializers.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def update(self, instance, validated_data):
+        questions = validated_data.pop('questions')
+        for key, value in validated_data.items():
+            setattr(instance, key, value)
+        instance.save()
+
+        for question in questions:
+            options = question.pop('options') if 'options' in question else None
+            Question.objects.filter(id=question['id']).update(**question)
```
edvardvb: What happens if the filter returns None? Get instead of filter?

lego/apps/surveys/serializers.py
```diff
@@ -0,0 +1,160 @@
...
+
+            if options is not None:
+                for option in options:
+                    Option.objects.filter(id=option['id']).update(**option)
```
edvardvb: Same

lego/apps/surveys/tests/test_submissions_api.py
```diff
@@ -0,0 +1,97 @@
...
+        self.regular_user = User.objects.get(username='abakule')
+
+    def test_create_admin(self):
+        """Admin users should be able to create submissions"""
+        self.client.force_authenticate(user=self.admin_user)
+        response = self.client.post(_get_list_url(1), submission_data(self.admin_user))
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+    def test_create_attended(self):
+        """Users who attended the event should not be able to create submissions"""
```
edvardvb: Shouldn't this test that attended users *can* create submissions?

lego/apps/surveys/tests/test_submissions_api.py
```diff
@@ -0,0 +1,97 @@
...
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_create_regular(self):
+        """Regular users should not be able to create submissions"""
+        self.client.force_authenticate(user=self.regular_user)
+        response = self.client.post(_get_list_url(1), submission_data(self.regular_user))
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_detail_admin(self):
+        """Admin users should be able to see detailed submissions"""
```
edvardvb: Should they? Are admins allowed to know what each attendee answered for each question and not just a summary/statistics?

lego/apps/users/fixtures/test_abakus_groups.yaml
edvardvb: Yep

lego/apps/surveys/serializers.py
```diff
@@ -90,17 +90,11 @@ def create(self, validated_data):
...
-                                    len(answer['selected_options']) > 1:
-                        raise exceptions.ValidationError('You cannot select multiple options for '
-                                                         'this type of question.')
-
-                    Answer.create(submission, question, **answer)
-                else:
-                    new_answer = Answer.objects.create(submission=submission, question=question,
-                                                       **answer)
-                    new_answer.save()
+                if getattr(question, 'question_type') is 1 and len(answer['selected_options']) > 1:
```
edvardvb: Any reason for this `getattr` and not just `question.question_type is 1...`

lego/apps/surveys/serializers.py
finnss: https://docs.djangoproject.com/en/1.9/topics/db/queries/#updating-multiple-objects-at-once
The `update` function is defined to work on a QuerySet, not a single object. The "appropriate" way to do it would be something like
```
q = Question.objects.get(...)
q.someValue = ...
q.otherValue = ...
q.save()
```

But that seem tedious, and [this StackOverflow post](https://stackoverflow.com/questions/2712682/django-update-object) seems to indicate that `update` is fine (or even better) for single objects too.

edvardvb: :+1:

lego/apps/surveys/serializers.py
finnss: I believe that's what I did initially, but I think `question` at this point is some sort of Django manager or pointer or some such, and iirc `question.question_type` resulted in something like `Error: Django.RelatedObjectManager has no attribute "question_type"`.

lego/apps/surveys/tests/test_submissions_api.py
finnss: I was using Google Forms as a reference and they provide full access to the creators of forms, including individual answers, so that's what I did initially. Might want to display the submission but without username in the frontend, I guess.

lego/apps/surveys/tests/test_submissions_api.py
finnss: Ah yeah ups, I must have copypasted from the regular user test and forgot to fix.

lego/apps/surveys/serializers.py
finnss: Blaming you when this causes an issue 🙃

lego/apps/surveys/tests/test_submissions_api.py
edvardvb: I think it's important that submissions are 'anonymous' for those reviewing them. It might be  a bit scary to give feedback on an event or Bedkom itself when you know that someone in Bedkom can read your name next to your comment.

lego/apps/users/fixtures/test_abakus_groups.yaml
finnss: Remind me!

lego/apps/surveys/permissions.py
```diff
@@ -0,0 +1,46 @@
...
+        elif view.action in ['retrieve']:
+            return user_attended_event and \
+                   survey.submissions.get(id=view.kwargs['pk']).user_id is user.id
```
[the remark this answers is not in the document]
finnss: Probably causes a crash, yea. I suppose providing a pk without a corresponding Submission should return 404, but from permissions I guess the best we can do is returning `False` and end up serving the user a 403?

odinuge:
If you are having problems with yapf, you can use [this](https://hub.docker.com/r/odinuge/yapf/) docker image. :smile: 

Execute this inside the root of the lego repo (the last arg is the path you will format)
```bash
$ docker run --rm -it -v $PWD:/code/ odinuge/yapf yapf -ir lego
```

PR #847, opened 2017-10-19
[WIP] Change payment_due_date to payment_due_days

hanswilw:
We want to give give each user X days to pay after they register. Arrkom has issues with people not paying, and we use this to warn and unregister people.


hanswilw:
This branch might be of interest when fixing #1088.

PR #881, opened 2017-10-23
Update fixtures and update serializer in joblistings

tmsolber:


lego/apps/joblistings/serializer.py
```diff
@@ -60,6 +62,11 @@ def update(self, instance, validated_data):
             new_workplaces.append(workplace)
         instance.workplaces.set(new_workplaces)
 
-        instance.save()
+        if responsible:
```
ekmartin: can't you just do `instance.responsible = validated_date.pop('responsible', None)` here? or can `responsible` sometimes be a falsy value that you want to change to `None`? 

actually, why do you need this at all? i might be missing something, but shouldn't responsible be set by `super().update(instance, validated_data)`?

tmsolber: @ekmartin  We had a problem that when the field was not set and we sent null to the backend the responsible field was not updated to null. When you save the form without a chosen responsible the old responsible was not removed. I don't know why...  but that is why we did this, and it works.

ekmartin: Maybe you were sending `undefined` and not `null`? When updates are done with `PATCH`, only the fields that are included are changed (so that you don't have to send all fields when changing something).

With the explicit `responsible` handling you have here you'll have to send `responsible` in every payload, even if you're doing a `PATCH` (and then see if it works after you remove the logic here).

Maybe add a test case to `test_joblistings_api` that checks that the responsible is cleared if you send `'responsible': None`?

eirsyl: Isn't `allow_null=True` enough here?

PR #893, opened 2017-10-24
Fix meetings queryset

odinuge:
Disable keyword-permissions in meetings. We can instead add an
admin-settings at a later stage.


eirsyl:
I think we should have some tests that verify this change...  ;p

odinuge:
Yeah. This is broke, because `/sudo/admin/meetings/create` and `/sudo/admin/meetingInvitations/create` are required...

odinuge:
@eirsyl This filters the queryset as we do to generate the ical.

PR #899, opened 2017-10-25
Make user able to join abakus in usersettings

hanswilw:


lego/apps/users/models.py
```diff
@@ -299,8 +300,17 @@ def email_address(self):
     def profile_picture(self, value):
         self.picture = value
 
+    @PermissionsMixin.is_abakus_member.setter
+    def is_abakus_member(self, add_to_group):
```
ekmartin: This is pretty implicit - that changing a boolean on a user causes changes to many to many relations. I guess the alternative would be to have a separate endpoint, or use the existing endpoint for adding memberships to groups - but not sure if it's better.

If this is the best solution it should at least include a few comments describing what it's doing. Maybe the `add_to_group` variable should be called something like `new_value` instead, to signify that it's the value that the user is trying to set `is_abakus_member` to (if I understand it correctly).

Also - where's the tests?

lego/apps/users/models.py
```diff
@@ -299,8 +300,17 @@ def email_address(self):
...
+        if self.is_verified_student():
```
ekmartin: what happens if the user isn't a verified student? it just ignores it? seems like this should either be validated somewhere else, or fail loudly here

lego/apps/users/models.py
```diff
@@ -299,8 +300,17 @@ def email_address(self):
...
+            abakus_group = AbakusGroup.objects.get(name=constants.MEMBER_GROUP)
+            if add_to_group:
+                abakus_group.add_user(self)
+            else:
+                abakus_group.remove_user(self)
+
     def is_verified_student(self):
-        return self.student_username is not None
+        return self.student_username is not None and self.grade is not None
```
ekmartin: shouldn't these be equivalent? to me it seems like if we have to check both of these here, we're not doing validation properly somewhere else - i.e. should you be able to be in a grade without having a student username?

lego/apps/users/models.py
```diff
@@ -160,7 +160,8 @@ def is_superuser(self):
 
     @property
     def is_abakus_member(self):
-        return 'Abakus' in [group.name for group in self.all_groups]
+        return constants.MEMBER_GROUP in [group.name for group in self.all_groups]
```
ekmartin: 👍 

not that it's relevant to this PR, but this seems a bit inefficient in general. do we have to retrieve all groups to figure out if one of them is abakus? maybe there's a more efficient way to figure that out with MPTT. 

if the reasoning is that `self.all_groups` is a cached property and this always results in a cache hit, we should at least use a regular for loop to avoid having to store the whole list of group names in memory (and then going through it again to do the `in` lookup): 

```python
for group in self.all_groups:
    if group.name == constants.MEMBER_GROUP:
        return True

return False
```

lego/apps/users/models.py
hanswilw: Well, it is implicit - but I felt it natural to do the group handling while updating the user. I mean since we are already passing a `is_abakus_member` boolean. I thought of the using the membership endpoint as you mention, but that would require a bit more logic frontend – while this would just be another radiobutton in the user settings page.

I should have added a WIP to this PR – I wanted feedback on this approach before finishing it up with tests.

lego/apps/users/models.py
hanswilw: It should fail hard 😊. The main point of checking if the user is verified is to only allow active students to join Abakus, not regular Users.

We could do this validation in the serializer, like we do with usernames, and raise eventual exception there instead if that is preferred.

lego/apps/users/models.py
hanswilw: They should, but the fixtures are not - will fix that and revert this. But we have to take this into consideration when adding bumping of grades, since we will have to remove the student_username - since this check should be false when they are alumni.

PR #906, opened 2017-10-26
[WIP] Open event list and view for unauthenticated users

edvardvb:
Quick idea of how to allow unauthenticated users to list and view events, but not see the list of attendees. Breaks the webapp hard.

pls discuss


eirsyl:
Shouldn't object permissions handle this? Try to set public=True on events.

eirsyl:
ObjectPermissions isn't implemented in the frontend, maybe trash it and remove the complexity :p

edvardvb:
@eirsyl @ek @warlo 
This opens events for everyone, but the "trick" used to hide the names of registered users from companies is to check for keyword permissions. This was done to avoid a lot of queries, and for active students this works fine, since students have the 'view'-permission.

However, for events that are open for alumni, this won't work. They will be able to register, but won't be able to see the names of others that have registered, since the 'Abakus' group doesn't have the 'view'-permission. Thoughts?

What are your thoughts about rewriting the permission system (other than "we should do it")? Should we just get this fixed and look at the system later?

The tests are failing because they test that unauthorized users can't see events, which this PR is changing of course.

odinuge:
Any progress on this @edvardvb? (ref. discussion here: https://github.com/webkom/lego-webapp/pull/875)

If this is far away from done, imo. we can open the metadata (for SEO) for everyone. I think it will benefit Abakus a lot, because sharing links on facebook looks terrible atm..

Maybe these values are enough to start with?

- Image
- Title
- Description

To simplify we can return these values from a custom endpoint, and use it only when the user isn't authenticated. Any thoughts?

edvardvb:
@odinuge As of right now this "works", but the tests should be rewritten before merging. There's a lot to do on the front end though. Also, it might be an idea to change this to open the number of waiting registrations to unauthenticated users as well.

There's still the issue of ex-students. Events where they are allowed to register are open for them to see and join, but the list of people who have registered isn't. Ideally we'd use the permission groups on pools to decide whether or not you can see other attendees, but I think that's going to be quite heavy on the queries? Also, we need a way to hide certain events from those who can't join them, such as events that are only for Abakom. I'd like to look into this tomorrow though!

I agree with you, those values should be enough to start with and will improve link sharing a lot!

PR #918, opened 2017-10-30
Allow for unsubscription to the weekly mail

edvardvb:
Would something like this work, @eirsyl? Don't really know how to test this :confused:


eirsyl:
works for now i guess 😄

edvardvb:
Merge so we can test in prod? :innocent:

eirsyl:
Maybe add a test that creates a notification setting for a user and makes sure lookup_recipients dont return the email?

lego/apps/restricted/tests/test_models.py
```diff
@@ -37,3 +40,31 @@ def test_invalid_lookup(self):
...
+                                           enabled=True,
+                                           channels=[PUSH])
+        NotificationSetting.objects.create(user=subscribed,
+                                           notification_type=WEEKLY_MAIL,
+                                           enabled=True,
+                                           channels=[EMAIL])
+
+        recipients = restricted_mail.lookup_recipients()
+
+        self.assertCountEqual(recipients, [
```
eirsyl: Assert list equal, not count.

PR #926, opened 2017-11-02
Event celery logging

hanswilw:
Turns out our celery beat tasks aren't running correctly – which is quite bad since they do a lot of important tasks. 

I was able to get an error when running celery beat locally, where the tasks were not sent due to `logger_context` being passed to tasks without `base=AbakusTask` – which does not make any sense. Though I cant seem to be able to reproduce that error by reverting these changes, so it might be something else. We should add these logs anyway though.


edvardvb:
this shit is so nice i cant even believe it

PR #944, opened 2017-11-07
Fix counter incrementing using an old counter value on unregister

hanswilw:
https://sentry.abakus.no/webkom/lego/issues/3549/events/1992473/

The counter was incrementing an old value when people unregistered. Meaning that a full event with e.g. 9 users, was incremented to 10 on unregister with bump from waiting list. This solves the problem by fetching the pool after the unregister happened, and using pool_id property to check if it exists to avoid another query.


ekmartin:
👍

PR #967, opened 2017-11-15
Yapf

odinuge:
https://github.com/google/yapf


README.md
```diff
@@ -52,6 +52,16 @@ foo = long_function_name(
     var_two
 )
 ```
+## Yapf - TODO
```
eirsyl: More generic section, add information about isort and flake8 to?

odinuge: Smart :+1: Would be nice with a few links etc.

Will look at it tomorrow. Yapf has a few settings (in `setup.cfg`), and I have tried to configure it more or less equally to how the codebase is formatted today (based on pep8). 

It is eg. nice that if you add a comma (`,`) in the end of a list/tuple/set, it will split all elements to separate lines (and the opposite if you don't).

odinuge:
Looks like `drone` works now, with the forced docker image pull... 

Thoughts anyone?

README.md
```diff
@@ -27,30 +27,24 @@ $ python manage.py runserver
 
 ## Code Style
 
-This codebase uses the PEP8 code style. We enforces this with flake8. We use almost all the rules
-used in the Requests project.
-
-In addition to the standards outlined in PEP 8, we have a few guidelines:
+This codebase uses the PEP8 code style. We enforces this with isort, yapf & flake8.
```
ekmartin: `We enforce` :)

ekmartin: PEP8 or PEP 8? let's choose one

ekmartin:
Sounds like a good idea to me :) Honestly, I wouldn't mind having no configured options for yapf except for PEP8 either (except for max line length 100), even if it means changes to how the code currently looks, but @eirsyl can decide.

README.md
```diff
@@ -27,30 +27,24 @@ $ python manage.py runserver
...
+In addition to the standards outlined in PEP 8, we have a few guidelines
+(see `setup.cfg` for more info):
 
 * Line-length can exceed 79 characters, to 100, when convenient.
 * Always use single-quoted strings (e.g. 'lego'), unless a single-quote occurs within the string.
 
-Additionally, one of the styles that PEP8 recommends for line continuations completely lacks all
-sense of taste, and is not to be permitted within the Requests codebase:
-
+Format the code with yapf in an editor plugin, or run:
```
ekmartin: Is there a decent pycharm plugin for yapf? https://plugins.jetbrains.com/plugin/9705-yapf-pycharm says it needs yapf globally installed, but haven't tested. Could be nice since a lot of people use pycharm, and it's what we recommend to new members.

README.md
odinuge: Guess I should read the text i change.. :upside_down_face:

README.md
odinuge: A few links would be nice! :smile: 
Haven't tested pycharm yet, but in vim it just werkz :tm:

odinuge:
@ekmartin Yeah; but I think these make sense:

- `split_before_named_assigns = false` when this is true (default) in breaks flake8
- `split_arguments_when_comma_terminated = true` - Allows us to write nicer tuples/lists
```python
  tup = ( 
      'abcsa',
      'sadasdasd'
      'saddwww', // Notice the last comma here
  )
  tup = ('abcsa', 'sadasdasd' 'saddwww')
```

- `dedent_closing_brackets = true`
```python
  # Gives this
  run_function(
     dsadadsa="dsadasdsadsad", sdadasd="wadasd", wdqqd="wqdqwdqwdwq",      
     dsasdasdsad="ewdw"
  )
  # instead of this
  run_function(dsadadsa="dsadasdsadsad", sdadasd="wadasd", wwqd="wqdqwdqwdwq",
               dsasdasdsad="ewdw")
```

Any thoughts @eirsyl ?

hanswilw:
`dedent_closing_brackets` should be true indeed, makes it enforce the style already noted in the readme 👍

PR #972, opened 2017-11-16
Add endpoint for searching for a registration for an event

martinhath:
10/10 code thank you come again

Neither me nor @orhanhenrik wasn't too sure about what to do about the permission logic. Also, should we return the user object, so that we can get a nice "Velkommen, Gunnar Bovim" to the frontend?

### DIFF
--- a/lego/apps/events/views.py
+++ b/lego/apps/events/views.py
@@ -21,12 +21,14 @@ from lego.apps.events.serializers.registrations import (AdminRegistrationCreateA
                                                         RegistrationPaymentReadSerializer,
                                                         RegistrationReadDetailedSerializer,
                                                         RegistrationReadSerializer,
-                                                        StripeTokenSerializer, RegistrationSearchSerializer)
+                                                        RegistrationSearchSerializer,
+                                                        StripeTokenSerializer)
 from lego.apps.events.tasks import (async_payment, async_register, async_unregister,
                                     check_for_bump_on_pool_creation_or_expansion,
                                     registration_payment_save)
 from lego.apps.permissions.api.views import AllowedPermissionsMixin
 from lego.apps.permissions.utils import get_permission_handler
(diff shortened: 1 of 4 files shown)


lego/apps/events/views.py
```diff
@@ -201,3 +204,32 @@ def admin_register(self, request, *args, **kwargs):
...
+
+    def create(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        username = serializer.data['username']
+
+        try:
+            reg = self.get_queryset().get(user__username=username)
+        except Registration.DoesNotExist:
+            raise ValidationError({'error': 'No user with that username'})
```
ekmartin: not that we use the error anyway, but maybe this should say something like `User bob is not registered for the event`?

ekmartin:
LGTM otherwise (@eirsyl?)

lego/apps/events/views.py
```diff
@@ -201,3 +204,32 @@ def admin_register(self, request, *args, **kwargs):
...
+
+        if not get_permission_handler(Event).has_perm(request.user, 'EDIT', obj=reg.event):
+            raise PermissionDenied()
+
+        reg.presence = constants.PRESENT
+        reg.save()
+        data = PublicUserSerializer(reg.user).data
```
ekmartin: actually, @martinhath, maybe this should return the updated registration? think that would work better with the frontend and reducers and shit (we have the full name from the search result anyway)

martinhath: Ah, right! Yeah, that sounds like a good idea.

lego/apps/events/views.py
martinhath: Too much typing. Close as WONTFIX

PR #975, opened 2017-11-16
Raise error if a user tries to register event precense twice

martinhath:


lego/apps/events/views.py
```diff
@@ -230,6 +230,9 @@ def create(self, request, *args, **kwargs):
                 'error': f'User {reg.user.username} is not registered for this event'
             })
 
+        if reg.presence != constants.UNKNOWN:
+            raise ValidationError({'error': f'User {reg.user.username} is already present.' })
```
ekmartin: Maybe we should send out an error code with these errors as well? I started writing code to handle them specifically in the frontend, and handling the string feels a bit ugly. We probably want to do something like this on a more global level later, but could just do something like `'error_code': 'already_present'` and `'error_code': 'not_registered'` to get started?

lego/apps/events/views.py
```diff
@@ -220,14 +221,29 @@ def create(self, request, *args, **kwargs):
...
+            })
+
         try:
-            reg = self.get_queryset().get(user__username=username)
+            reg = self.get_queryset().get(user=user)
         except Registration.DoesNotExist:
-            raise ValidationError({'error': 'No user with that username'})
+            raise ValidationError({
+                'error': 'The registration does not exist',
+                'error_code': 'not_registered'
```
hanswilw: Make these codes constants? :)

martinhath: Yeeeeah, we could, but it seemed kind of premature, so I figured we could just inline them. 

In the long run we might want to define a set of errors, shared between the back and front-end such that we can handle errors correctly on the front-end without weird string parsing stuff.

PR #976, opened 2017-11-17
Handle decode error when parsing restricted email attachments

eirsyl:


eirsyl:
Deployed this to production, IDI need this to send an email to users attending to Julebord.

PR #1017, opened 2018-01-22
Make sure users are bumped even when `heed_penalties` is False

edvardvb:


lego/apps/events/models.py
```diff
@@ -431,20 +431,19 @@ def pop_from_waiting_list(self, to_pool=None):
...
                         registration.user,
                         [to_pool], penalties
                     )
-                if self.heed_penalties and penalties < 3 and earliest_reg < timezone.now():
-                    for group in registration.user.all_groups:
-                        if group in permission_groups:
+                    if penalties < 3 and earliest_reg < timezone.now():
+                        if self.can_register(registration.user, to_pool):
                             return registration
+                else:
```
ekmartin: can merge the `else` and the `if` into an `elif self.can_register(..)` :)

PR #1027, opened 2018-01-30
Permission Wars III: Revenge of the Serializers

edvardvb:
This pull request makes the assumption that all existing events will get `can_view_groups=None` (this is already the case in prod, but not in the local fixtures), and that keyword permissions for view and list are removed from regular students. Instead, events will be open for everyone, and certain events can be hidden using the `is_abakom_only`-property. This property sets the Abakom group as a `can_view_group`, and filters the queryset accordingly. 

The list of registrations will be hidden unless you can actually register for the event (or you're part of abakom or creator of the event, for admin purposes). Instead, the registration field returns the number of registrations of that pool.

Will require som frontend work :innocent:


lego/apps/events/serializers/events.py
```diff
@@ -79,6 +80,23 @@ class Meta:
         )
         read_only = True
 
+    def user_should_see_regs(self, event, user):
```
edvardvb: Maybe it's possible to move this to the view and pass it to the serializer, so we won't have to call it twice?

edvardvb:
ffs `require_auth=False` allows anyone to edit the event this is rip

edvardvb:
Someone merge this, I don't want to look at it anymore :weary:

lego/apps/events/serializers/pools.py
```diff
@@ -28,9 +28,14 @@ def create(self, validated_data):
...
-            return RegistrationPaymentReadSerializer(queryset, context=self.context, many=True).data
-        return RegistrationReadSerializer(queryset, context=self.context, many=True).data
+        should_see_regs = self.context.get('should_see_regs', None)
+        if should_see_regs:
+            if obj.event.is_priced:
+                return RegistrationPaymentReadSerializer(
+                    queryset, context=self.context, many=True
+                ).data
+            return RegistrationReadSerializer(queryset, context=self.context, many=True).data
+        return -1
```
edvardvb: This can be changed to return something more meaningful, such as number of attendees, but I think this is enough for now

lego/apps/events/permissions.py
```diff
@@ -9,6 +10,56 @@
...
     perms_without_object = [CREATE, 'administrate']
+    authentication_map = {VIEW: False, LIST: False}
+    force_queryset_filtering = True
+
+    def filter_queryset(self, user, queryset, **kwargs):
+        if not user.is_authenticated or not user.is_abakom_member:
+            return queryset.filter(can_view_groups=None)
+        return queryset
+
+    def has_perm(
```
hanswilw: This is duplicating the `permissions/permissions.py` quite extensively, I think its better to do the abakom check explicitly and do super().has_perm to reuse the default logic. 

Maybe something like this is better?
```
can_view_event = True
if obj and isinstance(obj, ObjectPermissionModel) and obj.is_abakom_only:
   can_view_event = user.is_abakom_member
return can_view_event and super().has_perm(*args, **kwargs)
```

lego/apps/events/models.py
```diff
@@ -518,6 +518,11 @@ def spots_left_for_user(self, user):
 
         return sum([pool.spots_left() for pool in pools])
 
+    @property
+    def is_abakom_only(self):
```
hanswilw: I see your point here, making the event only visible for abakom if abakom is in can_view_groups - but couldnt we just have a `is_abakom` boolean flag on the event model instead? That way we wont rely on the can_view_groups and we can filter directly on it. Have you thought of that?

lego/apps/events/serializers/events.py
```diff
@@ -79,6 +80,27 @@ class Meta:
...
+        return PoolReadSerializer(queryset, context=self.context, many=True).data
+
+    def get_waiting_registrations(self, obj):
+        request = self.context.get('request', None)
+        if request and \
+                request.user.is_authenticated and \
+                self.user_should_see_regs(obj, request.user):
+            queryset = obj.waiting_registrations
+            return RegistrationReadSerializer(queryset, context=self.context, many=True).data
+        return obj.waiting_registrations.count()
```
odinuge: Is it possible to return this count as `registrationCount` instead? Same for waiting list..

Would make typings in the frontend way better.. :smile:

hanswilw: Yep, just add `registration_count` as we do [here](https://github.com/webkom/lego/blob/ccd839e9ee61508f49afff487df7259cfdd029ab/lego/apps/events/serializers/events.py#L54) to the detailed serializer @edvardvb

hanswilw: And drop these `.count()` here.

odinuge: Hmm. If you remove count here, it will return all the registrations for unauthorized users.

It is still nice to get pools with registration counts, not just the total count.

lego/apps/events/serializers/events.py
hanswilw: I think we are better off handling this in the viewset, because this check is done for every pool in the event. This could be solved by just switching serializers. One without pools and registrations, and one with. That way you can just inherit the one without pools and add the pool and waiting_registration fields to minimize duplicate code.

hanswilw: Or wait, i think it should only run once - how did you get this to run twice? Anyways, better off selecting different serializers imo.

edvardvb: It'll run during `get_pools` to pass an argument to the pool serializer about whether or not to include registrations, and one during `get_waiting_registrations` to check whether to include all waiting regs or to just return the number of waiting regs

edvardvb: Would it be ok to include `waiting_registrations_count` and `pool.registration_count` for every event, or only those where registrations are hidden?

lego/apps/events/serializers/events.py
hanswilw: SerializerMethodField does not return anything implicitly, does it?

Right, I was only thinking of the total - like we use for the calendar. We should probably support showing the pools with separate counters on the event page. Could add a registrationCount property, similar to the one for event, for the pools and add it in the pool serializer.

hanswilw: Oh, I see what you meant. I meant to drop the whole line ofc :)

lego/apps/events/serializers/events.py
```diff
@@ -74,7 +74,7 @@ class Meta:
             'comment_target', 'start_time', 'end_time', 'merge_time', 'pools',
             'unregistration_deadline', 'company', 'active_capacity', 'feedback_description',
             'feedback_required', 'is_priced', 'price_member', 'price_guest', 'use_stripe',
-            'payment_due_date', 'use_captcha', 'waiting_registrations', 'tags', 'is_merged',
+            'payment_due_date', 'use_captcha', 'waiting_registration_count', 'tags', 'is_merged',
```
odinuge: This serializer also needs `is_abakom_only`, in order to retain the value when opening the eventEditor :smile:

edvardvb:
@hanswilw @odinuge final thoughts?

eirsyl:
Todo:
* Remove custom permissions functions
* Switch serializers based on permissions
* Add serializer flagg called `is_abakom_only `
 -> TRUE => {require_auth=True, can_view_groups:[Abakom]}
 -> FALSE => {require_auth=False, can_view_groups:[]}

edvardvb:
@hanswilw

lego/apps/events/views.py
```diff
@@ -57,12 +57,21 @@ def get_queryset(self):
             queryset = Event.objects.all()
         return queryset
 
+    def user_should_see_regs(self, event, user):
+        return event.get_possible_pools(user, future=True, is_admitted=False).exists() or \
```
hanswilw: Did you check how this affects number of queries?

edvardvb: Not sure if my numbers are correct, but I got the same number of queries (35) both when running this and when running master. Retrieving when not authenticated results in 11 queries. It sounds odd that the numbers are equal, so I'm not sure what to believe.

lego/apps/events/serializers/events.py
```diff
@@ -111,29 +120,38 @@ class EventCreateAndUpdateSerializer(TagSerializerMixin, BasisModelSerializer):
...
+        is_abakom_only = validated_data.pop('is_abakom_only', False)
         with transaction.atomic():
             event = super().create(validated_data)
             for pool in pools:
                 permission_groups = pool.pop('permission_groups')
                 created_pool = Pool.objects.create(event=event, **pool)
                 created_pool.permission_groups.set(permission_groups)
-
+            if is_abakom_only:
+                event.require_auth = True
```
hanswilw: Would be nice to have this as a model method. Then you can just call e.g. `event.set_abakom_only()` both here and in update

lego/apps/events/serializers/events.py
```diff
@@ -96,6 +96,14 @@ def get_price(self, obj):
             return obj.get_price(user=request.user)
 
 
+class EventReadAuthUserDetailedSerializer(EventReadUserDetailedSerializer):
```
hanswilw: Nitpicking, but maybe PublicEventSerializer and PublicEventDetailedSerializer could be more descriptive, instead of adding auth in these?

edvardvb: Hmm..

We have the following serializers:
`EventRead` for list,
`EventReadDetailed` for read, but no user info or registrations
`EventReadUserDetailed` for read and user info, but no registrations
`EventReadAuthUserDetailed` for read, user info and registrations.

Maybe rename to List, Read, ReadUser and ReadDetailed?

odinuge:
Together with https://github.com/webkom/lego-webapp/pull/1098, this should work now!  :smile: 


#1062 Should fix registration/websockets, by revering the ish/non-working v2 version of `django-channels`. I works well locally, so it should act equally in PROD too. But, did you @eirsyl  do any changes to staging? Or should everything work when the new channels version is reverted?

eirsyl:
The deployment pipeline is updated for channels 2, everything breaks :trollface:

PR #1063, opened 2018-02-11
Fix ical permissions

odinuge:
Also add a abakom-test


odinuge:
This is more or less just a partial revert of 64f8a169481f169ed06703a5af44b8f5afea9b9f, with a few new test cases.

PR #1073, opened 2018-02-13
Start parallelizing tests

ekmartin:
This makes it possible to run most tests (not the Cassandra ones) in parallel by running: `python manage.py test --parallel --exclude-tag feed`. This makes the 637 non-feed tests run in ~35 seconds, instead of 100 seconds on my machine. To be able to run the rest of the tests in parallel as well we'll have to figure out how to isolate the different test processes to different Cassandra key spaces.

It does so by adding a few new base test classes that mock out Cassandra, and forces tests that want to hit Cassandra to inherit `FeedTestBase`. This also stops a bunch of tests that didn't need to hit Cassandra from doing so.

It also disables migrations in tests, which speeds up the initial test load time.

### DIFF
--- a/lego/apps/ical/tests/test_api.py
+++ b/lego/apps/ical/tests/test_api.py
@@ -143,7 +143,7 @@ class IcalAuthenticationTestCase(BaseAPITestCase):
             self.assertEqual(res.status_code, 401)
 
 
-class IcalPersonalTestCase(BaseAPITestCase:
+class IcalPersonalTestCase(BaseAPITestCase):
     fixtures = [
         'test_abakus_groups.yaml',
         'test_users.yaml',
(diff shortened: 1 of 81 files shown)


lego/apps/ical/tests/test_api.py
```diff
@@ -143,7 +143,7 @@ def test_get_ical_with_invalid_token(self, *args):
             self.assertEqual(res.status_code, 401)
 
 
-class IcalPersonalTestCase(APITestCase):
+class IcalPersonalTestCase(BaseAPITestCase:
```
odinuge: Smaaal syntax-error :sunglasses:

ekmartin: lolol, merge conflict fixing

PR #1074, opened 2018-02-14
Cache grade

hanswilw:
> There are 2 hard problems in computer science: cache invalidation and naming things. -- Phil Karlton

What TTL should we set? I just set a year, since this basically only changes once a year during bump. We could reduce it, but it would not fix anything - I guess.

Using event from fixtures with 9 registrations:
image

⬇️

image


ekmartin:
This might be useful for places where we retrieve only one user and don't want to join their class memberships (are there any places like that though?) - however I'm not convinced it actually solves the issue of the event admin page. 

Sure, you decrease the number of SQL queries in the current solution, but you replace each one of them with a query to Redis instead, right? Which is definitely faster, but still not free at all. It might be better to just do *one* expensive event admin query, e.g. by joining with all the class groups and looping through the result in Python to figure out which year someone belongs to (might even be able to do that in a query, but maybe not worth the complexity).

In NERD we cached all of a user's groups in general, mostly because of permissions if I remember correctly. This probably also isn't a good solution for places where you need more than one user, but might greatly improve the performance of e.g. search. We should definitely still try to avoid this for as long as possible though.

hanswilw:
Closing in favor of #1085

PR #1081, opened 2018-02-16
[WIP] Sort event attendees by relevance to you

finnss:
Currently generates more queries than the rest of LEGO combined, as well as probably being implemented nowhere near the correct place. But hey, you gotta start somewhere. The sorting algorithm itself works, at least.


finnss:
Unless I'm mistaken, number of queries is now `2n + 2` where `n` is the number of registrations. So for an event with 100 users attending, this sorting would generate 202 queries atm. Still way too high, but you gotta know your enemy, no?

The algorithm is called a bunch of times when accessing event detail, so I'm convinced it's implemented the wrong place.

hanswilw:
Closing in favor of #1086

PR #1085, opened 2018-02-17
Join grade queries

hanswilw:
What about this @ekmartin ?

This will only introduce one query by querying all grades and inject the grade to the registrations in serialized event dict with python. I had some takes on doing this in the django orm, but that is not trivial to join on prefetched values. 

Should probably add some tests and move the logic outside this function, but lets discuss the logic at first :)

### DIFF
--- a/lego/apps/events/views.py
+++ b/lego/apps/events/views.py
@@ -33,7 +33,6 @@ from lego.apps.permissions.api.views import AllowedPermissionsMixin
 from lego.apps.permissions.utils import get_permission_handler
 from lego.apps.users.constants import GROUP_GRADE
 from lego.apps.users.models import AbakusGroup, User
-from lego.apps.users.serializers.abakus_groups import AbakusGroupNameSerializer
 from lego.utils.functions import verify_captcha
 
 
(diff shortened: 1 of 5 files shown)


ekmartin:
Yeah, if it brings down the number of queries it's 👍

lego/apps/events/views.py
```diff
@@ -97,13 +99,29 @@ def perform_update(self, serializer):
...
+                'user'
+            ).prefetch_related('user__abakus_groups')),
             Prefetch('registrations', queryset=Registration.objects.select_related('user')),
         )
-        serializer = self.get_serializer(queryset.first())
-        return Response(serializer.data)
+        event = queryset.first()
+        event_data = self.get_serializer(event).data
+
+        grades = AbakusGroup.objects.filter(type=GROUP_GRADE)
```
ekmartin: if you do `grades = AbakusGroup.objects.filter(type=GROUP_GRADE).values('id', 'name')` you'll get a list of dicts with `{'id: 1, 'name': 'webkom'}` (instead of using the serializer)

hanswilw: Right, thats better 👍🏻

lego/apps/users/serializers/abakus_groups.py
```diff
@@ -43,3 +43,9 @@ class Meta:
             'logo',
             'type',
         )
+
+
+class AbakusGroupNameSerializer(serializers.ModelSerializer):
```
ekmartin: don't need this serialize anymore right?

hanswilw: I'll remove it when I refactor and add tests later 👍

PR #1086, opened 2018-02-17
Sort event users in django orm

hanswilw:
Changed a little bit of your sorting logic into a query 😉

What I am basically doing here is that I am `.annotate`-ing (adding) a field onto the queryset called `shared_memberships` which is constructed by `Count()`-ing the number of `user_abakus_groups` in the `current_user_groups`. Using a [`Q()` expression](https://docs.djangoproject.com/en/2.0/ref/models/querysets/#query-related-tools)

image
⬇️
image

You can consider merging this and adding some tests if you want to continue working on this feature 👍🏻


hanswilw:
@eirsyl Wanna check this out?

PR #1094, opened 2018-02-19
Add penalties API end-point

amalieeh:


lego/apps/users/permissions.py
```diff
@@ -98,3 +98,7 @@ def has_perm(
                 return data.get('user') == user.id
 
         return False
+
```
eirsyl: This class isn't used 😄 You can remove it

PR #1103, opened 2018-02-21
Add api endpoints for fetching survey templates

finnss:


lego/apps/surveys/views.py
```diff
@@ -22,7 +25,27 @@ def get_serializer_class(self):
...
+
+    @list_route(methods=['GET'])
+    def templates(self, request):
+        queryset = Survey.objects.all().prefetch_related('questions').filter(
+            template_type__isnull=False
+        )
+        serializer = self.get_serializer(queryset, many=True)
+        return Response(serializer.data)
+
+    @list_route(
```
ekmartin: should these possibly be in their own `TemplateViewSet` with list and retrieve? might be a bit cleaner. you can still get the URLs you're using by doing something like this: https://github.com/webkom/lego/blob/master/lego/api/v1.py#L139

lego/apps/surveys/views.py
```diff
@@ -12,8 +13,10 @@
 
 
 class SurveyViewSet(AllowedPermissionsMixin, viewsets.ModelViewSet):
-    queryset = Survey.objects.all().prefetch_related('questions', 'submissions')
+    queryset = Survey.objects.all().prefetch_related('questions', 'submissions'
+                                                     ).exclude(template_type__isnull=False)
```
ekmartin: why does it exclude surveys without templates here?

lego/apps/surveys/views.py
finnss: Yeah you're right, that would be cleaner. Will fix

ekmartin: could also just put the template viewset directly on the API root, i.e. `/api/v1/survey-templates` instead of `/api/v1/surveys/templates` - up to you

lego/apps/surveys/views.py
finnss: A survey having a `template_type` means that it _is_ a template, not that it's based on one. The idea was that fetching regular surveys and fetching templates were two separate things. However, this exclude means that templates cannot be seamlessly treated as regular surveys when you want to edit or delete or whatever, which was the motivation for not just making a separate Template model. So I'll remove this. I have to filter away templates in the survey list view front end anyway.

finnss:
I'm not really sure what all those `filter_backends` do, I added them at the behest of @eirsyl. I also kinda would like for the url to be `surveys/templates` instead of `survey-templates`, but I couldn't figure out how to make it work with a separate ViewSet for templates instead of `@list-route`. Doesn't really matter though. 

This seems to be working as intended now.

finnss:
I added some tests for templates 🙌

lego/apps/surveys/fixtures/survey_template_fixtures.yaml
```diff
@@ -0,0 +1,240 @@
+- model: surveys.Survey
+  pk: 1
+  fields:
+    title: 'Template: company_presentation'
+    event: 1
+    template_type: 'company_presentation'
```
finnss: I intend to load these template fixtures in prod, ala `./manage.py loaddata lego/apps/surveys/fixtures/survey_template_fixtures.yaml`. That's fine, right?

lego/apps/surveys/permissions.py
```diff
@@ -33,6 +40,8 @@ def has_permission(self, request, view):
             event=event.id, user=user.id, presence=constants.PRESENT
         ).exists()
 
+        if view.action in ['update', 'partial_update']:
+            return False
```
finnss: This means admins can't edit Submissions, which seems like a fine personal integrity measure to me.

finnss:
@eirsyl did you have some comments about this or is it good to go?

edvardvb:
Looks pretty good to me, but I think @eirsyl should take a second look at the permissions and any comments you had about the serializers :)

PR #1105, opened 2018-02-21
Pass along shared_memberships instead of ordering by it

hanswilw:


lego/apps/events/views.py
```diff
@@ -67,7 +67,7 @@ def get_registrations(self, user):
             query |= Q(user__abakus_groups=group)
         registrations = Registration.objects.select_related('user').annotate(
             shared_memberships=Count('user__abakus_groups', filter=query)
-        ).order_by('-shared_memberships', 'user_id')
+        )
         return registrations
```
finnss: This way we end up sorting the registrations _as well as_ passing the number of shared memberships, could that be an issue? As things stand, this would still break the waiting list, right? We discussed solving that by adding `prospective_pool` to registrations and sorting front-end, which seems like a good solution to me. Just want to note that merging this PR by itself would keep the breaking of waiting lists.

odinuge: This will just fetch calculate the `shared_memberships`, and not sort by them.

hanswilw: The point was to sort frontend only - the order by was therefore removed. 😊

PR #1112, opened 2018-02-23
Event responsible group and group contact email, fixes #1104

### LINKED ISSUE #1104: Event responsible group field
Link events to group. This makes it easier for us to direct the end user to the actual responsible persons.

eirsyl:
Fixes #1104


lego/apps/events/models.py
```diff
@@ -47,6 +47,9 @@ class Event(Content, BasisModel, ObjectPermissionsModel):
     company = models.ForeignKey(
         Company, related_name='events', null=True, on_delete=models.SET_NULL
     )
+    responsible_group = models.ForeignKey(
+        'users.AbakusGroup', on_delete=models.CASCADE, null=True, related_name='events'
```
ekmartin: shouldn't this be `SET_NULL`?

lego/apps/events/views.py
```diff
@@ -48,7 +48,9 @@ def get_queryset(self):
                 'pools', 'pools__registrations', 'company', 'tags'
             )
         elif self.action == 'retrieve':
-            queryset = Event.objects.prefetch_related('pools', 'pools__permission_groups', 'tags')
+            queryset = Event.objects.prefetch_related(
+                'pools', 'pools__permission_groups', 'tags', 'company', 'responsible_group'
```
ekmartin: shouldn't `responsible_group` be in a `select_related` since it's on the event model? and not a many to many field

eirsyl: Yeah my bad, to fast

lego/apps/events/serializers/events.py
```diff
@@ -64,6 +65,9 @@ class EventReadDetailedSerializer(TagSerializerMixin, BasisModelSerializer):
     comment_target = CharField(read_only=True)
     cover = ImageField(required=False, options={'height': 500})
     company = CompanyField(queryset=Company.objects.all())
+    responsible_group = AbakusGroupField(
+        queryset=AbakusGroup.objects.all(), required=False, allow_null=True
```
ekmartin: think we should change this to `required=True` in the serializer eventually

ekmartin:
good stuff

lego/apps/events/serializers/events.py
eirsyl: yes, only required in the webapp for right now

PR #1118, opened 2018-02-25
Fix event shared membership tests

odinuge:
The old test was not deterministic, and had a fail rate of 10-20%. It
depended on the order of the result, and the new one does not.

### DIFF
--- a/lego/apps/events/tests/test_events_api.py
+++ b/lego/apps/events/tests/test_events_api.py
@@ -270,7 +270,7 @@ class RetrieveEventsTestCase(BaseAPITestCase):
                     self.assertIsNone(reg['feedback'])
                     self.assertIsNone(reg['chargeStatus'])
 
-    def test_event_registration_sort_order_default(self):
+    def test_event_registration_no_shared_memberships(self):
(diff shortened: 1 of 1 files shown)


lego/apps/events/tests/test_events_api.py
```diff
@@ -289,17 +289,29 @@ def test_event_registration_sort_order_rearranged(self):
...
-            reg['sharedMemberships'] for reg in event_response.data['pools'][0]['registrations']
-        ]
-        self.assertEqual(memberships, [1, 0, 2])  # reg1 = 1, reg2 = 0, reg3 = 2
+
+        registrations = event_response.data['pools'][0]['registrations']
+
+        for reg in registrations:
+            reg_id = reg['id']
+            self.assertEqual(
+                reg['sharedMemberships'], shared_memberships_count[reg_id],
```
hanswilw: Hmm, I assumed the registrations from the payload was sorted by the id in addition to the registration date, but I see that it is only sorted by the registration date. We could add ['registration_date', 'id'] to the ordering on the model to fix this, but I suppose that there isnt a problem with not having any specific ordering on identical registration_date's :)

PR #1120, opened 2018-02-26
Survey mail

finnss:
Adds a celery task that sends an email to all users who attended a survey's event once `survey.active_from` is in the past.


lego/apps/surveys/tasks.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@celery_app.task(serializer='json', bind=True, base=AbakusTask)
+def send_survey_mail(self, logger_context=None):
+    self.setup_logger(logger_context)
+
+    surveys = Survey.objects.filter(active_from__lte=timezone.now(), sent=False)
+    print('surveys in send_mail', surveys.all())
+    for survey in surveys.all():
+        with transaction.atomic():
```
ekmartin: I don't think you need this transaction wrapper. Since you're always sending out all the notifications *before* you set `sent = True` and save, it'll only save if you managed to send out all the surveys (so the transaction doesn't matter at all). 

What might happen regardless is that it can create notifications for half the users, then crash, and the next time it tries to do so it'll send out notifications for those users again - but the transaction doesn't help with that (you'd need one `sent` field per user).

(also remove the prints)

lego/apps/surveys/templates/surveys/email/survey.html
```diff
@@ -0,0 +1,40 @@
...
+
+    <tr>
+        <td class="content-block">
+            Du har en ny undersøkelse å svare på for arrangement {{ event }}.
+        </td>
+    </tr>
+
+    <tr>
+        <td class="content-block">
+            <i>Å ikke svare på undersøkelsen innen fristen vil medføre en prikk.</i>
```
ekmartin: don't have to change this now maybe, but are we removing this when users can't join new events if they've got pending surveys?

lego/apps/surveys/tests/test_surveys_task.py
```diff
@@ -0,0 +1,25 @@
...
+    def setUp(self):
+        self.surveys = Survey.objects.all()
+
+    def test_sent_true(self):
+        active_surveys = self.surveys.filter(active_from__lte=timezone.now())
+        for survey in active_surveys:
+            self.assertEqual(survey.sent, False)
+
+        send_survey_mail()
+        sent_surveys = self.surveys.filter(active_from__lte=timezone.now())
```
ekmartin: instead of doing the filtering twice you can just loop over `active_surveys` again and call `survey.refresh_from_db()` before the assert (that way you're sure that you're checking the same items as well).

Should maybe add an assert to make sure that `active_surveys` contain X amount of items as well, so it's not empty?

lego/settings/celery.py
```diff
@@ -93,6 +93,10 @@ def on_setup_logging(**kwargs):
     'check-that-pool-counters-match-registration-number': {
         'task': 'lego.apps.events.tasks.check_that_pool_counters_match_registration_number',
         'schedule': crontab(hour='*', minute=0)
+    },
+    'notify_user_about_new_survey': {
+        'task': 'lego.apps.surveys.tasks.survey_mail',
+        'schedule': crontab(hour='*', minute=0)
```
ekmartin: might as well ramp this up to like every tenth minute or so, it's not like it does a lot of work anyway, and that way you don't have to wait an hour if you're unlucky with the survey start time (`minute="*/10"`)

ekmartin:
looks good :)

lego/apps/surveys/notifications.py
```diff
@@ -0,0 +1,32 @@
...
+
+        return self._delay_mail(
+            to_email=self.user.email,
+            context={
+                'name': self.user.full_name,
+                'survey': survey.id,
+                'event': survey.event.title,
+                'active_from': survey.active_from,
+            },
+            subject=f'Undersøkelse etter Abakus-arrangement',
```
edvardvb: Nitpicking, but maybe `f'Spørreundersøkelse for {survey.event.title}'` or something like that. Could be weird with certain event-names, but I think having the event title in the subject makes sense.

lego/apps/surveys/templates/surveys/email/survey.html
```diff
@@ -0,0 +1,40 @@
+{% extends "email/base.html" %}
+
+{% block alert %}
+
+    <tr>
+        <td class="alert alert-warning">
+            Du har en ny undersøkelse å svare på
```
edvardvb: "Vennligst svar på denne spørreundersøkelsen fra Abakus" or similar?

lego/apps/surveys/templates/surveys/email/survey.html
edvardvb: Yup. I think it would be a good idea to just implement this functionality straight away, shouldn't be too complicated :)

lego/apps/surveys/tasks.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@celery_app.task(serializer='json', bind=True, base=AbakusTask)
+def send_survey_mail(self, logger_context=None):
+    self.setup_logger(logger_context)
+
+    surveys = Survey.objects.filter(active_from__lte=timezone.now(), sent=False)
+    for survey in surveys.all():
+        for registration in survey.event.registrations.filter(presence=PRESENT):
+            if not Submission.objects.filter(survey=survey, user=registration.user).exists():
```
finnss: How about something like this? It will still notify a user several times if they don't have time to answer the survey in-between, but it should at least prevent any "I already answered this survey, why am I getting mail?!?!" messages.

ekmartin: Since you're setting `sent = True` it will only run once anyway though, right? That is, it won't send out any extra reminders since after the first time `sent` will be `True`.

I'm not sure if periodic reminders are actually necessary though, since they'll have to answer the survey before their next event regardless. And in that case, you don't need the `Submission.objects.filter` logic, since no one will be able to answer the survey before it's sent out.

finnss: I was thinking of this as a fix to the issue you described where the system could crash after some users had received a mail but before `sent` was set to `True`. Like I said, this would only help if the users who receive mails before the crash have time to answer the survey before its sent again, but that's better than nothing?

lego/apps/surveys/templates/surveys/email/survey.html
finnss: Just rejecting the request to register backend would indeed not be too hard. However, we would also have to return something sensible to the front-end (list of unanswered surveys? Full objects or just ids?), then reduce that properly into the redux state (which would normally expect a Registration object), and finally display a proper error message to the end user. None of that is necessarily difficult, but it would take time to implement. Do we want to delay surveys launch for this functionality, if it comes to that?

Ofc feel free to just do it and submit a pr 🙃

lego/apps/surveys/tasks.py
ekmartin: Ah, I see - I misunderstood then. I don't think that's a problem you need to solve, if it somehow crashes in the middle of your task (which would involve something crazy like the database doing down) I don't think it really matters if people get an extra email once.

finnss: Alright, cool! Will drop this check then

Issue #1039, opened 2018-02-02
Meeting invitation email missing "reject" button


odinuge:
The meeting invitation email is missing a "reject"/"cannot attend" button. The link is present in the clear text version of the notification, so the proper link can be borrowed from there.

cherrybeans:
It's a feature, not a fault :+1: We don't want people rejecting!!

PR #1125, opened 2018-02-27
New Mail template

finnss:
From
skjermbilde 2018-02-27 kl 18 08 27

To
skjermbilde 2018-02-27 kl 18 06 54

I'm very open to feedback on the design. Should the text be left-aligned? Different colors? Bigger text?
 Any input is welcome. 

I have yet to go through all the different app templates (like meeting) and update them to work with the new version, I figured it would be best to agree that this is a good idea first.

I also haven't been able to test the new template in different mail clients. I tested the template by generating `.eml` files and opening them with the OSX mail viewer, but it seems like most online mail clients like Outlook and Gmail don't have support for opening such files. I'm not sure how to go about testing in those clients. 

NB: Currently changes a config file that fucks up mail-sending (so I could test), so don't merge with that in there.


odinuge:
Nice!
Pls fix #1039 a 😎😄👌🎉

SmithPeder:
Looks very good to me. I prefer to always press the "Les mer " link at the bottom, and think that should be the "main" thing to press in the invitation. So that the instant "accept and decline" should be sub-alternatives.

finnss:
Got this tested in the most common mail clients, see below. The Logo is not shown in all of them because it's only hosted locally atm.

Gmail
skjermbilde 2018-03-05 kl 16 23 36

Outlook
skjermbilde 2018-03-05 kl 16 23 48

Inbox
skjermbilde 2018-03-05 kl 16 25 19

finnss:
I think we should make the links more look link-y, and the text should be slightly darker for increased contrast with the background.

eirsyl:
Why not replace the base email template and use this version for all emails lego produces?

finnss:
I did?

The syntax is slightly different for creating buttons in the new template, so I'll have to update every template that used them. I initially only updated the meeting template so I had an example to work with.

eirsyl:
ye, that was the point other email may look ugly now? Is it also possible to ignore the smtp server changes in development.py?

eirsyl:
Its better to serve the files from abakus.no rather than s3.amazonaws.com 😄 The hardcoded logo served from 127.0.0.1 require users to run LEGO on their own machine, i think its best to avoid that :trollface: 

Nice template btw, i'm all in for using this!

finnss:
Yeah ofc, like I said at the meeting I figured we should agree to switch to this new template before I spent time on properly configuring and updating everything. But it sounds like people like it so I guess we're good to go:)

BTW because I'll be gone for xcom in the upcoming month I guess it might be a good idea for someone else to take over this project. Or we could wait ofc

lego/templates/email/base.html
```diff
@@ -1,310 +1,273 @@
...
+              <v:textbox inset="0,0,0,0">
+            <![endif]-->
+              <center>
+                <table cellpadding="0" cellspacing="0" width="600" class="w320">
+                  <tr>
+                    <td class="pull-left mobile-header-padding-left" style="vertical-align: middle;">
+                      <a href=""><img width="150" height="30" src="http://127.0.0.1:8000/static/assets/img/logo.png" alt="logo"></a>
+                    </td>
+                    <td class="pull-right mobile-header-padding-right" style="color: #4d4d4d;">
+                      <a href="https://www.facebook.com/groups/398146436914007/"><img width="38" height="47" src="http://s3.amazonaws.com/swu-filepicker/LMPMj7JSRoCWypAvzaN3_social_09.gif" alt="facebook" /></a>
```
edvardvb: This should link to the official facebook page for Abakus, not the internal group. Also, it would be nice to have links to Instagram (and LinkedIn?) as well :)

PR #1144, opened 2018-03-12
Fix penalty when unregistering from waiting list

hanswilw:
Penalties were not given to users that unregistered from waiting list, which is not the intended behaviour decided by Bedkom and HS.


edvardvb:
Uhm, are you sure? I thought the rule was that you were given a penalty even if you're _bumped_ after the unregistration limit, not if you're still in the waiting list?

hanswilw:
Heh, you are right. Not sure what I was thinking, we should fix the frontend though - since it shows that you get a penalty when unregistering while on waiting list.

edvardvb:
Yeah, that makes sense! :)

PR #1146, opened 2018-03-12
[WIP] Social auth feide backend

eirsyl:
For student verification.

- [ ] Require login before trying to verify account with feide
- [ ] Redirect to a status page after feide falidation
- [ ] Act based on UserSocialAuth.extra_data.groups


odinuge:
Any more plans for this (other than the loginflow)? I did a test, and it looks awesome!

The only "relevant" group i could find was MTDT (MTKOM for komtek):
```json
    {                                                                                                                                                         
        "displayName": "Datateknologi - masterstudium (5-årig)",                                                                                         
        "id": "fc:fs:fs:prg:ntnu.no:MTDT",                                                                                                                    
        "membership": {                                                                                                                                       
            "active": true,                                                                                                                                   
            "basic": "member",                                                                                                                                
            "displayName": "Student",                                                                                                                         
            "fsroles": [                                                                                                                                      
                "STUDENT"                                                                                                                                     
            ]                                                                                                                                                 
        },                                                                                                                                                    
        "parent": "fc:org:ntnu.no",                                                                                                                           
        "type": "fc:fs:prg",                                                                                                                                  
        "url": "http://www.ntnu.no/studier/mtdt"                                                                                                              
    },
```

We still need a clever way to determine the grade/year. The best solution I can think of is to just remove all grades each summer, and then bump (using `Membershiphistory`) users to the next year/grade when they verify that they are still active students. We just have to make sure we don't kick people from mailinglists etc.

odinuge:
I have rebased this branch to master, so everything should be ready for some work.

To make this work we need to rewrite the authentication-, verification- and/or login flow. Before starting the implementation, we need to agree about how we should do it. Thoughts anyone??

cc: @JonasBak

JonasBak:
Do we know anything about how this works in between semesters, as this is probably when we want to bump people? I mean before students have registered for the semester and payed the semester fee, are they still registered as active students? Also, how should we handle edge cases like students transferring between studies in the start/middle of semesters?
I'm going to have to take a better look and read up on some stuff to be able to say anything about how we should handle flow.

PR #1151, opened 2018-03-13
Allow users to check if they have answered surveys, as intended

finnss:
Should fix https://github.com/webkom/lego-webapp/issues/1164


eirsyl:
Tests?

finnss:
@eirsyl

PR #1156, opened 2018-03-15
Pin boto3 to version 1.6

ekmartin:
Since our tests never hit AWS we're never gonna see test failures from non-breaking changes to boto3, so might as well leave it pinned to a slightly more relaxed version to save us some upgrade noise.


eirsyl:
Have been waiting on this huge patch for quite some time 😄

PR #1184, opened 2018-04-10
[Ignore for now] Add a /mine endpoint to surveys

finnss:
Currently returns all the surveys the requesting user has created, which means it's only relevant for admin users. Makes sense atm because regular users can't fetch the list endpoint. When/if we change that, this endpoint could be changed to returning all surveys the user has access to (ie all surveys for events they've attended, which usually equals all surveys they've answered)

Implemented frontend in https://github.com/webkom/lego-webapp/pull/1203


eirsyl:
This can be solved with filters, create a FilterSet

PR #1193, opened 2018-04-22
Survey token

finnss:
Adds a new endpoint, `/survey-results/`. This is a detail-only endpoint, and it _requires_ a token to be supplied for access. That means the only valid uses of this endpoint is on the form `/survey-results/<pk>/?token=<token>`. When used in this way, the endpoint doesn't check any permissions except for the validity of the token. This means anyone can see the results of a survey if they have the token, which lets us share the results with the companies for which the survey is made.

The token itself is a new field in the Survey model, and is a randomly generated 64 character long string unique to each survey. The idea is that this token is _only_ served to admins, who get a nice button frontend that uses the token to make a shareable link on the form given above.


lego/apps/surveys/authentication.py
```diff
@@ -0,0 +1,22 @@
...
+
+        raw_token = request.GET.get('token')
+        if not raw_token:
+            return
+
+        try:
+            Survey.objects.get(token=raw_token)
+            return User.objects.get(
+                pk=1
+            ), raw_token  # TODO: Select a prod user for this or solve differently
```
finnss: Gonna need some feedback on this. `authenticate` needs to return a tuple of (user, token) for authentication to be successful, but what user should be associated with people trying to access survey results by using a token? This `pk=1` referred to a test user without any permissions in my testing, and that worked fine. Is it an acceptable solution to make an anonymous user for this purpose and use his pk here? Or can we somehow work around this problem some other way? @eirsyl @hanswilw @odinuge

lego/apps/surveys/views.py
```diff
@@ -52,6 +55,8 @@ class SubmissionViewSet(AllowedPermissionsMixin, viewsets.ModelViewSet):
     filter_class = SubmissionFilterSet
     filter_backends = (DjangoFilterBackend, )
 
+    # authentication_classes = api_settings.DEFAULT_AUTHENTICATION_CLASSES + [SurveyAuthentication]
```
finnss: Don't I need this here as well? Seems to work without, but I don't understand how

PR #1195, opened 2018-04-24
Add tags to the event serializer for surveys

finnss:
Prevents crash when navigating from surveys to events


odinuge:
Smooth. We should make some kind of "base serializers" for all entities, and make flowtypes according to them. That would help us a lot with these kind of issue.

We don't always need tags, and therefore we can use the current flow type for events:
```
type Event = {
    ...stuff,
    tags?: Array<Tag>
}
```

PR #1204, opened 2018-04-26
Survey hide answers

finnss:
Based on #1193, merge that first

Adds a new field to the Answer model, `hide_from_public`. This can be set to True or False using a separate endpoint, `/surveys/<pk>/submissions/<submission_pk>/hide/?answer=<answer_pk>`. Swap `hide` for `show` for the counterpart.

When this field is set to `True`, that answer is no longer included in the aggregated data used in the public, token-based `/survey-results/`. Thus, admin users get the option to hide sensitive or inappropriate comments before displaying the results of a survey to a company.


lego/apps/surveys/views.py
```diff
@@ -68,3 +74,45 @@ def create(self, request, *args, **kwargs):
...
+    @detail_route(methods=['GET'])
+    def show(self, request, **kwargs):
+        submission, answer = self.validate_answer(request, **kwargs)
+        answer.show()
+        return Response(
+            data=SubmissionReadSerializer(submission).data, status=status.HTTP_202_ACCEPTED
+        )
+
+
+class SurveyTokenViewset(viewsets.GenericViewSet):
```
odinuge: Since `token` in Survey is `unique=True`, do you need a `viewset` then? There should be no need to supply the pk of the survey. The only thing used is `GET` on a single endpoint (more or less), so a simple [view](http://www.django-rest-framework.org/api-guide/views/) should be more then enough, or?

lego/apps/surveys/views.py
```diff
@@ -68,3 +74,45 @@ def create(self, request, *args, **kwargs):
         return Response(
             SubmissionReadSerializer(serializer.instance).data, status=status.HTTP_201_CREATED
         )
+
+    def validate_answer(self, request, **kwargs):
```
odinuge: Imo `validate_answer` does not describe what this piece of code does.

Something like:
```python
def resolve_answer(self, request, **kwargs):
```
And then a method call on the object.

If there is any logic about what question types can be hidden, it should live inside the model.

lego/apps/surveys/serializers.py
```diff
@@ -46,7 +46,9 @@ class AnswerSerializer(BasisModelSerializer):
 
     class Meta:
         model = Answer
-        fields = ('id', 'submission', 'question', 'answer_text', 'selected_options')
+        fields = (
+            'id', 'submission', 'question', 'answer_text', 'selected_options', 'hide_from_public'
```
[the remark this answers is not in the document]
odinuge: And the owner/sender of the `submission` should not be able to see `hide_from_public` field.

lego/apps/surveys/models.py
```diff
@@ -49,10 +81,19 @@ class Answer(BasisModel):
...
 
     def create(submission, question, **kwargs):
         selected_options = kwargs.pop('selected_options')
         answer = Answer.objects.create(submission=submission, question=question, **kwargs)
         for selected_option in selected_options:
             answer.selected_options.add(selected_option)
         answer.save()
+
+    def hide(self):
+        self.hide_from_public = True
```
odinuge: The logic can live inside here! :smile:

lego/apps/surveys/models.py
```diff
@@ -15,6 +22,31 @@ class Survey(BasisModel):
...
+                text_answers = list(
+                    map(
+                        lambda answer: answer.answer_text,
+                        Answer.objects.filter(question=question).exclude(hide_from_public=True)
+                    )
+                )
+                shuffle(text_answers)
+                options['answers'] = text_answers
+            else:
+                for option in question.options.all():
```
odinuge: I think this can be done inside the database using aggregation.

lego/apps/surveys/models.py
```diff
@@ -15,6 +22,31 @@ class Survey(BasisModel):
     )
     event = models.OneToOneField('events.Event', on_delete=models.CASCADE)
     sent = models.BooleanField(default=False)
+    token = models.CharField(max_length=64, default=generate_new_token, unique=True)
+
+    def aggregate_submissions(self):
```
odinuge: We should (not that important yet) write a few simple tests on this function. Would be a lot easier to refactor without manual testing then :slightly_smiling_face:

lego/apps/surveys/models.py
finnss: What logic?

PR #1207, opened 2018-05-03
Meeting comment

cherrybeans:


cherrybeans:
Tested it locally, and it works as far as I can see. Mobile responsiveness could be improved with nested comments though.

PR #1208, opened 2018-05-23
Temporary disable pagination for survey submissions

odinuge:
Temporary fix for this (about 80 submissions).

screenshot from 2018-05-23 13-39-00

We should instead use the aggregated view, and maybe use infinite scroll on the different questions. No big deal tho, becuase of the small amount of data :smile:


odinuge:
LGTM

PR #1233, opened 2018-06-14
Add filter for joblistings by company

### LINKED ISSUE #913: Create filter by company in joblistings
Currently `https://apiserver/api/v1/joblistings/?company=id` returns all active joblistings

mrfarstad:
Closes #913 🎉

### DIFF
--- a/lego/apps/joblistings/filters.py
+++ b/lego/apps/joblistings/filters.py
@@ -1,0 +1,9 @@
+from django_filters import FilterSet
+
+from lego.apps.joblistings.models import Joblisting
+
+
+class JoblistingFilterSet(FilterSet):
+    class Meta:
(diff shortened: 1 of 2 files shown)


odinuge:
Nais :+1: 

It may be possible to use a `Filterset` like we do in events:
https://github.com/webkom/lego/blob/master/lego/apps/events/filters.py#L9
https://github.com/webkom/lego/blob/master/lego/apps/events/views.py#L41

Something like:
```python
class JoblistingsFilterSet(FilterSet):
    company = CharFilter('company')

    class Meta:
        model = Joblisting
```

If that doesn't work, this will do the job :smile:

odinuge:
@mrfarstad Should we just close this?

PR #1249, opened 2018-07-29
Add user grade bump script

odinuge:
The script will bump users that has not been bumped in X amount of time.
It will also check if users have memberships in multiple student groups,
and warn about that.

Usage:
```bash
$ ./manage.py bump_users             # Check the status / "dry run"
$ ./manage.py bump_users --no-dryrun # The real deal
```

Should be more or less good to go, but a few tests would be nice. :smile:
- Done :heavy_check_mark:


lego/utils/management/commands/bump_users.py
```diff
@@ -0,0 +1,118 @@
...
+
+komtek_groups_names = [
+    f'1. klasse {KOMTEK_LONG}',
+    f'2. klasse {KOMTEK_LONG}',
+    f'3. klasse {KOMTEK_LONG}',
+    f'4. klasse {KOMTEK_LONG}',
+    f'5. klasse {KOMTEK_LONG}',
+]
+
+timedelta_since_prev_bump = timedelta(minutes=1)
```
odinuge: :trollface:

PR #1255, opened 2018-08-15
Support change grade on profile

SmithPeder:
Supports webkom/lego-webapp#1255


assets/test.json
```diff
@@ -0,0 +1,23 @@
+{
```
odinuge: :taco:

SmithPeder: Muy bien! hahaha

Issue #1272, opened 2018-09-11
When verifying as a student, do not join first grade if you are already in a higher grade


mrfarstad:
This is a bug where people end up being in multiple grades if you verify your account after being a member for more than one year

kotlarz:
Fixed in the referenced pull request

PR #1277, opened 2018-09-23
1272 student confirmation

kotlarz:
Needs PR #1276 merged in before this PR.

Fixes issue #1272.

### DIFF
--- a/lego/apps/users/views/student_confirmation.py
+++ b/lego/apps/users/views/student_confirmation.py
@@ -90,18 +90,22 @@ class StudentConfirmationPerformViewSet(viewsets.GenericViewSet):
         user.student_username = student_confirmation['student_username']
         course = student_confirmation['course'].lower()
 
+        has_grade_group = user.has_grade_group
+
         if course == constants.DATA:
             course_group = AbakusGroup.objects.get(name=constants.DATA_LONG)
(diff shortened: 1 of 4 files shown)


lego/apps/users/views/student_confirmation.py
```diff
@@ -90,18 +90,22 @@ def create(self, request, *args, **kwargs):
...
             course_group = AbakusGroup.objects.get(name=constants.DATA_LONG)
             course_group.add_user(user)
 
-            grade_group = AbakusGroup.objects.get(name=constants.FIRST_GRADE_DATA)
-            grade_group.add_user(user)
+            if not has_grade_group:
+                grade_group = AbakusGroup.objects.get(name=constants.FIRST_GRADE_DATA)
+                grade_group.add_user(user)
         else:
             course_group = AbakusGroup.objects.get(name=constants.KOMTEK_LONG)
```
odinuge: You can remove this line too! (and the same for `DATA_LONG`), since the users isn't supposed to be members of `Datateknologi` or `Kommunikasjonsteknologi` (becuse of the tree structure, uses will get the proper permissions from being in a grade group).

It is also not supported by the bump script: https://github.com/webkom/lego/pull/1249

Issue #1266, opened 2018-09-01
Event should be marked as favorite when signing up for this event


liseph:

kotlarz:
We don't have favorite function like we did in NERD as far as I know? The "star" next to the even title is only used to _pin_ events to the front page.

edvardvb:
I dont think the star does anything right now. The backend supports following/favoriting events afaik

kotlarz:
My bad, seems like we do indeed have a follow function (for users and events too!), I just got too caught on the "favorite" name instead of "follow". I'll get this fixed within today :smile:

SmithPeder:
Yehh I think so

PR #1281, opened 2018-10-01
Make the user follow/unfollow events when they register/unregister for that event.

kotlarz:
Also works on admin register/unregister. Fixes issue #1266


odinuge:
This fixes #938

PR #1287, opened 2018-10-11
[WIP] Add support for bulk admin register

odinuge:


lego/apps/events/views.py
```diff
@@ -240,12 +240,23 @@ def admin_register(self, request, *args, **kwargs):
...
-            registration = event.admin_register(**serializer.validated_data)
+            with transaction.atomic():
+                for user in users:
+                    try:
+                        registration = event.admin_register(**serializer.validated_data, user=user)
+                    except RegistrationExists:
+                        error_users.append(user)
+
+                if len(error_users) != 0:
+                    raise APIRegistrationsExists(users=error_users)
```
orhanhenrik: Should add some tests for this behavior.

lego/apps/events/views.py
```diff
@@ -240,12 +240,23 @@ def admin_register(self, request, *args, **kwargs):
             raise APIEventNotFound()
         serializer = self.get_serializer(data=request.data)
         serializer.is_valid(raise_exception=True)
+
+        users = serializer.validated_data.pop('users')
```
orhanhenrik: Maybe we should support passing in a single user as well? Then there is no "downtime" during deployment since the api will be backwards compatible.

PR #1296, opened 2018-10-19
Hide memberships of deleted groups

odinuge:
This fixes https://sentry.abakus.no/webkom/lego-webapp/issues/6016/?query=is:unresolved, since all memberships will now have a valid group.

### DIFF
--- a/lego/apps/users/models.py
+++ b/lego/apps/users/models.py
@@ -194,20 +194,24 @@ class PermissionsMixin(CachedModel):
     @abakus_cached_property
     def memberships(self):
         return Membership.objects.filter(
-            deleted=False,
+            abakus_group__deleted=False,
             is_active=True,
             user=self,
         )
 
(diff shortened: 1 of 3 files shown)


eirsyl:
seems legit :trollface: 👍

lego/apps/users/models.py
```diff
@@ -194,20 +194,24 @@ class Meta:
     @abakus_cached_property
     def memberships(self):
         return Membership.objects.filter(
-            deleted=False,
```
hanswilw: Is it intended to remove this deleted kwarg, but still have it in`all_groups` below?

odinuge: Ahh, guess it can be removed from `all_groups` as well. `deleted=False` is implicit because the default model manager is `PersistentModelManager`, that only returns the items with `deleted=False`

PR #1301, opened 2018-10-24
Companies overhaul

LudvigHz:


lego/apps/companies/views.py
```diff
@@ -31,6 +31,7 @@ class CompanyViewSet(
 ):
     queryset = Company.objects.all()
     permission_classes = (permissions.IsAuthenticated, )
+    ordering = 'name'
```
odinuge: Smart :+1:

lego/apps/companies/serializers.py
```diff
@@ -79,10 +80,21 @@ class CompanyListSerializer(BasisModelSerializer):
         }
     )
 
+    event_count = serializers.SerializerMethodField()
+
+    def get_event_count(self, obj):
```
odinuge: We generally have all the class values before functions in our serielizer, so the line `joblisting_count = serializers.SerializerMethodField()` should be over this function :smile:

lego/apps/companies/serializers.py
```diff
@@ -79,10 +80,21 @@ class CompanyListSerializer(BasisModelSerializer):
...
+        return obj.events.filter(start_time__gt=timezone.now()).count()
+
+    joblisting_count = serializers.SerializerMethodField()
+
+    def get_joblisting_count(self, obj):
```
odinuge: Have you checked how many `sql`-queries these two method fields generates? I suspect they will create two queries per company, and that would result in `1 + (2 * companyCount)` queries (if all the prefetching is done in the view). :thinking:

lego/apps/companies/serializers.py
LudvigHz: got it! Will fix

lego/apps/companies/serializers.py
LudvigHz: As far as I can tell, they only generate 1 query each, not sure how to properly check this though.

PR #1302, opened 2018-10-25
[WIP] Improve pinning

JonasBak:
The current pinning system is pretty bad.

Here are some ideas to make it better:
 * Pin events and articles for groups
 * Pin events and articles between two dates
 * Give a better overview of which items are pinned and when

Things that need to be done:
 * Tie it together with frontpage
 * Permissions :100: 
 * Tests :1st_place_medal: 
 * Frontend :rice_ball:


lego/apps/pinned/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
```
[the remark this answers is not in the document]
SmithPeder: :)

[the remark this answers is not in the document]
JonasBak: ez django-admin startapp :kissing_heart:

lego/apps/pinned/models.py
```diff
@@ -0,0 +1,19 @@
...
+from lego.apps.users.models import AbakusGroup
+from lego.utils.models import BasisModel
+
+
+class Pinned(BasisModel):
+    pinned_from = models.DateField()
+    pinned_to = models.DateField()
+
+    event = models.ForeignKey(
+        Event, related_name='pins', on_delete=models.SET_NULL, null=True, blank=True
```
odinuge: Maybe cascade these, so we don't end up with `pinned` values without an event or article?

lego/apps/pinned/views.py
```diff
@@ -0,0 +1,34 @@
...
+
+class PinnedViewSet(AllowedPermissionsMixin, viewsets.ModelViewSet):
+
+    permission_classes = (IsAuthenticated, )
+    serializer_class = CreatePinnedSerializer
+    queryset = Pinned.objects.all()
+
+    @list_route(methods=['GET'])
+    def mine(self, request, pk=None):
+        queryset_events = Pinned.objects.filter(
```
odinuge: Have you tested how many queries these create?

lego/apps/pinned/views.py
```diff
@@ -0,0 +1,34 @@
...
+from lego.apps.permissions.api.views import AllowedPermissionsMixin
+from lego.apps.pinned.models import Pinned
+from lego.apps.pinned.serializers import (
+    CreatePinnedSerializer, PinnedArticleSerializer, PinnedEventSerializer
+)
+
+
+class PinnedViewSet(AllowedPermissionsMixin, viewsets.ModelViewSet):
+
+    permission_classes = (IsAuthenticated, )
```
odinuge: :trollface:

odinuge:
What is the status on this?? :man_shrugging: Quite far behind master already... :/

JonasBak:
> What is the status on this??  Quite far behind master already... :/

I got a bit stuck working on the frontend as i could not find any ways of showing what is pinned that actually improves the current system from a users perspective. The biggest issue with the current system is that when more than one item is pinned, only one of them shows up as pinned on the frontpage, and i found it hard to improve on this without adding too much clutter to the frontpage. I've talked a bit with others about how to solve this, but will probably need more help.

As for the conflicts; none of the conflicting files are actually that important, so I don't think it will be that much work rebasing.

PR #1316, opened 2018-11-07
Polls

mrfarstad:


lego/apps/frontpage/views.py
```diff
@@ -52,12 +54,23 @@ def list(self, request):
                 request.user, queryset_events_base
             )
 
+        try:
```
odinuge: You can also do something like this, so you don't have to do `try-except` and the extra if.

```
poll_qs = Poll.objects.filter(pinned=True).order_by("created_at").last()
poll = DetailedPollSerializer(poll_qs, _context_stuff_).data if poll_qs else None
```

lego/apps/polls/serializers.py
```diff
@@ -0,0 +1,148 @@
...
+    comment_target = CharField(read_only=True)
+
+    options = OptionSerializer(many=True)
+    total_votes = IntegerField(read_only=True)
+
+    has_answered = serializers.SerializerMethodField()
+
+    def get_has_answered(self, obj):
+        user = self.context["request"].user
+        return user in obj.answered_users.all()
```
odinuge: Will this retrieve all users from db to compare in python? A better solution would be to use a proper query to check.

lego/apps/polls/tests/test_api_polls.py
```diff
@@ -0,0 +1,92 @@
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_vote_poll_authenticated_answered(self):
+        """A user should not be able to vote on a poll more than once"""
+        self.client.force_authenticate(self.authenticated_user)
+        self.client.post(_get_vote_url(2), {"optionId": 3})
+        response = self.client.post(_get_vote_url(2), {"optionId": 4})
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_vote_poll_unauthenticated(self):
```
odinuge: Maybe add some tests on the `valid_until` functionality?

lego/apps/users/fixtures/test_abakus_groups.yaml
```diff
@@ -173,3 +173,8 @@
   fields: {deleted: false, name: Students, description: '', contact_email: '', parent: null,
     logo: null, type: annen, text: '', permissions: '[]', show_badge: true, lft: 1,
     rght: 2, tree_id: 17, level: 0}
+- model: users.abakusgroup
```
odinuge: This file is generated by this file: https://github.com/webkom/lego/blob/474ae2536507e304a87e007e542e0710ba24b57e/lego/apps/users/fixtures/test_abakus_groups.py (as you see on the first few lines :smile:)

lego/apps/polls/models.py
```diff
@@ -0,0 +1,61 @@
...
+        return Option.objects.filter(poll=self).aggregate(total_votes=Sum("votes"))[
+            "total_votes"
+        ]
+
+    def get_absolute_url(self):
+        return f"{settings.FRONTEND_URL}/polls/{self.id}/"
+
+    def vote(self, user, option_id):
+        if user in self.answered_users.all():
+            return "forbidden"
```
odinuge: Instead of hardcoding this, it is better to use proper exceptions:

https://github.com/webkom/lego/blob/474ae2536507e304a87e007e542e0710ba24b57e/lego/apps/users/views/student_confirmation.py#L33

lego/apps/polls/models.py
```diff
@@ -0,0 +1,61 @@
...
+        option = self.options.get(pk=option_id)
+        if option:
```
odinuge: Wrap this inside an transaction maybe?

And what should be returned if `!option`?

lego/apps/polls/serializers.py
odinuge: This will also generate a query per poll, so when listing polls, this will be quite ineffective. You can instead aggregate the data when selecting the queryset

lego/apps/polls/views.py
```diff
@@ -0,0 +1,35 @@
...
+        elif self.action == "retrieve":
+            return DetailedPollSerializer
+        return PollSerializer
+
+    @decorators.detail_route(methods=["POST"], permission_classes=[IsAuthenticated])
+    def vote(self, request, *args, **kwargs):
+        poll = self.get_object()
+        vote_result = poll.vote(request.user, request.data["option_id"])
+        serializer = self.get_serializer_class()(poll, context={"request": request})
+        if vote_result == "forbidden":
```
odinuge: Same here: better with proper exceptions :smile:

lego/apps/frontpage/views.py
LudvigHz: Nice! Thats alot cleaner

lego/apps/polls/tests/test_api_polls.py
```diff
@@ -0,0 +1,98 @@
...
```
odinuge: Should this be `test_vote_poll_authenticated`?

lego/apps/polls/tests/test_api_polls.py
```diff
@@ -0,0 +1,98 @@
...
+        """A user with no permissions should be able to vote on a poll"""
+        self.client.force_authenticate(self.unauthenticated_user)
+        response = self.client.post(_get_vote_url(1), {"optionId": 1})
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_vote_poll_not_valid(self):
+        """A user should be able to vote on a poll where valid_until is not in the future"""
+        self.client.force_authenticate(self.authenticated_user)
+        response = self.client.post(_get_vote_url(3), {"optionId": 5})
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
```
odinuge: Maybe add a test for `test_edit_poll_without_permissions` (and create), with normal useres without permissions to edit/create polls?

lego/apps/polls/tests/test_api_polls.py
```diff
@@ -0,0 +1,98 @@
...
+        """A user with no permissions should be able to retrive the detailed poll view"""
+        self.client.force_authenticate(self.unauthenticated_user)
+        response = self.client.get(_get_detail_url(2))
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_vote_poll_authenticated_unanswered(self):
+        """A user with permissions should be able to vote on a poll that he has not answered"""
+        self.client.force_authenticate(self.authenticated_user)
+        response = self.client.post(_get_vote_url(2), {"optionId": 3})
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
odinuge: Maybe validate that the answered_count has increased?

lego/apps/polls/tests/test_api_polls.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def test_vote_poll_authenticated_answered(self):
+        """A user should not be able to vote on a poll more than once"""
+        self.client.force_authenticate(self.authenticated_user)
+        self.client.post(_get_vote_url(2), {"optionId": 3})
+        response = self.client.post(_get_vote_url(2), {"optionId": 4})
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
```
odinuge: Maybe validate that the answered_count has _not_ increased?

lego/apps/polls/tests/test_models.py
```diff
@@ -0,0 +1,15 @@
...
+class PollMethodTest(BaseTestCase):
+    fixtures = ["test_users.yaml", "test_polls.yaml"]
+
+    def setUp(self):
+        self.user = User.objects.get(username="test1")
+        self.poll = Poll.objects.get(pk=1)
+
+    def test_str(self):
+        title = "Test poll 1"
+        self.assertEqual(str(self.poll), title)
```
odinuge: Would be nice with some logic testing.

Eg. voting increases the count of total votes, and the count of users that has voted. And wise versa for the opposite.

lego/apps/polls/models.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def get_absolute_url(self):
+        return f"{settings.FRONTEND_URL}/polls/{self.id}/"
+
+    @transaction.atomic
+    def vote(self, user, option_id):
+        option = self.options.get(pk=option_id)
+        if not option:
+            raise ParseError(detail="Option not found.")
+        option.votes += 1
```
JonasBak: The votes are increased from the model without checking if the user has voted already, should we allow that or should we move that logic in here too?

lego/apps/polls/serializers.py
```diff
@@ -0,0 +1,151 @@
...
+    def update(self, instance, validated_data):
+        options = validated_data.pop("options")
+
+        # Update the regular fields that aren't options first
+        super().update(instance, validated_data)
+
+        # Delete options that aren't in the received list
+        for old_option in instance.options.all():
+            options_with_ids = filter(lambda q: "id" in q, options)
+            existing_options_ids = map(lambda q: q["id"], options_with_ids)
```
JonasBak: Could we move these two lines to before the loop as they don't need to be run each iteration?

lego/apps/polls/serializers.py
```diff
@@ -0,0 +1,151 @@
...
+            if old_option.id not in existing_options_ids:
+                old_option.delete()
+
+        # Add or update option, depending on the received option has an id
+        for option in options:
+            if "id" in option:
+                Option.objects.filter(id=option["id"]).update(**option)
+                Option.objects.get(id=option["id"])
```
JonasBak: Why do we `get` the object that we just updated?

lego/apps/polls/models.py
LudvigHz: It is checked in the view already, it is neccesary to verify it here as well?

JonasBak: I just wondered if there was any reason to have one check in this method (the option exists) and rely on that the user is checked elsewhere. Is there any reason not to check both here (and not in the view)? But then again, might be an anti-pattern I'm not aware of.

LudvigHz: I guess it is cleaner and safer to have all validation in the model.

lego/apps/polls/serializers.py
LudvigHz: yup

lego/apps/polls/tests/test_api_polls.py
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+    def test_create_poll_authenticated(self):
+        """A user with permissions should be able to create a poll"""
+        self.client.force_authenticate(self.authenticated_user)
+        response = self.client.post(_get_list_url(), self.poll_data)
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+    def test_create_poll_unathenticated(self):
+        """A user without persmissions should not be able to create a poll"""
```
cherrybeans: "permissions", not "permsmissions" :upside_down_face: xD

lego/apps/polls/models.py
```diff
@@ -0,0 +1,69 @@
...
+    return timezone.now() + timezone.timedelta(weeks=52)
+
+
+class Poll(Content, BasisModel):
+
+    description = models.TextField(blank=True)
+
+    valid_until = models.DateTimeField(default=get_time_delta)
+
+    answered_users = models.ManyToManyField(User, related_name="answered_polls")
```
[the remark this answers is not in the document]
odinuge: As long as voting is done in a trasaction, that should be no problem. The biggest reason to keep the current design is privacy, and imo. that wins. We should also do the same for surveys.

If we need more "data" on each vote (eg. grade from user), we should add that to some kind of separate table; to keep the privacy at a ok level.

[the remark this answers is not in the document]
orhanhenrik: I don't think a database transaction helps here, it will only ensure that operations are ACID.  
Since it is a poll I don't think anybody expects it to be anonymous, it's also quite common to be able to see what you answered in the result box. But I guess it's nice to have it anonymous even if you lose that one UI feature.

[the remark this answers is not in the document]
orhanhenrik: Ah okay I guess the many-to-many relation will crash if you try to insert a duplicate, in that case the transaction will be enough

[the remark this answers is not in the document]
odinuge: Ahh. Haven't thought about the fact that users may want to see what they voted for tho.. :thinking: That might bee smooth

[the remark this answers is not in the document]
odinuge: bee :bee:

lego/apps/polls/serializers.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+
+    @transaction.atomic
+    def update(self, instance, validated_data):
+        options = validated_data.pop("options")
+
+        # Update the regular fields that aren't options first
+        super().update(instance, validated_data)
+
+        # Delete options that aren't in the received list
```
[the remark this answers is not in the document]
LudvigHz: I think you should be able to remove an option if there has been a mistake. The votes will be lost, but I think that's fine. However, since there is no relation between the user and an option, the user will still be listed as answered. Which maybe isn't the best behavior, but there's no way to remove the user along with the vote. Thoughts?

PR #1346, opened 2019-01-13
Add postgres as tooltip

SmithPeder:


kotlarz:
https://github.com/webkom/lego/wiki/Noob-Guide

SmithPeder:
I think it should be present in the quick startup guide, as it's the only thing missing for a full setup 💯

PR #1349, opened 2019-01-21
Update copyright year

edvardvb:
🤓


SmithPeder:
It would be great if you implemented these changes in https://github.com/webkom/lego-webapp/blob/master/LICENSE

edvardvb:
@SmithPeder Thanks for the input, I'll see what I can do. Bit of a busy week though!

PR #1350, opened 2019-01-21
Only list companies marked as active

LudvigHz:


lego/apps/companies/serializers.py
```diff
@@ -112,6 +112,7 @@ class Meta:
             "address",
             "logo",
             "thumbnail",
+            "active",
```
edvardvb: Is this necessary? The frontend doesn't need to know whether the company is active or not, we just feed it a list of active companies? 

Or do we need it here to be able to filter the queryset in the view? I don't think so?

LudvigHz: Yeah, it's unnecessary. Removed it now.

PR #1351, opened 2019-01-21
Reduce the amount of queries on frontpage endpoint

odinuge:
Needs tests? Nah :man_shrugging: 

screenshot from 2019-01-21 23-10-54
screenshot from 2019-01-21 23-14-12


odinuge:
:trollface: 

We should write some tests when @SmithPeder finishes his refactoring in events.

PR #1352, opened 2019-01-23
Feedback fix

edvardvb:
Use `.__str__()` to ensure a string is passed for serialization, not the group itself.


edvardvb:
Uuuh, not sure why it brought in those two commits from Ludvig

edvardvb:
Tested in staging already, it produced no more issues in Sentry

PR #1354, opened 2019-01-24
[WIP] Improve event serializers

SmithPeder:
1) I wanted to improve the `EventReadSerializer` by removing the `company` field. We only use this serializer for the event list, and here the company field is not needed. This was creating more requests then needed. The `EventForSurveySerializer` however did extend the serializer so I moved company over there, as I do believe company is needed where that serializer is used.

2) The `EventSearchSerializer` is almost the exact same serializer as `EventReadSerializer`, except the `text`, `end_time`, `pinned` field. So I was wondering if we could just merge these two serializers into one as I've done here?

3) Start using the `EventReadSerializer` on the frontpage instead of the `EventSearchSerializer` like we do today.


odinuge:
@SmithPeder said this was abandoned.

PR #1365, opened 2019-02-08
Event ordering

LudvigHz:
This adds optional ordering to the event view via url params.

### DIFF
--- a/lego/apps/events/filters.py
+++ b/lego/apps/events/filters.py
@@ -1,4 +1,4 @@
-from django_filters import CharFilter, DateFilter, FilterSet
+from django_filters import CharFilter, DateFilter, FilterSet, OrderingFilter
 from lego.apps.events.models import Event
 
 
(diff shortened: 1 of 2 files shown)


lego/apps/events/filters.py
```diff
@@ -1,4 +1,4 @@
-from django_filters import CharFilter, DateFilter, FilterSet
+from django_filters import CharFilter, DateFilter, FilterSet, OrderingFilter
```
odinuge: Do you need this?

lego/apps/events/views.py
```diff
@@ -57,6 +59,12 @@
 class EventViewSet(AllowedPermissionsMixin, viewsets.ModelViewSet):
 
     filter_class = EventsFilterSet
+    filter_backends = (
+        DjangoFilterBackend,
+        filters.OrderingFilter,
+        LegoPermissionFilter,
+    )
+    ordering_fields = ("start_time", "end_time", "title", "registration_deadline_hours")
```
odinuge: I dont think we need `registration_deadline_hours`. It is only a fixed integer count of hours before the events starts, that the registration closes.

LudvigHz: I thought maybe you might want to use this in combination with the start time for sorting after when the registration closes. But maybe its not so useful?

lego/apps/events/filters.py
LudvigHz: nope, forgot to remove that

PR #1451, opened 2019-02-11
Add emailAddress to public user serializer

LudvigHz:


lego/apps/users/serializers/users.py
```diff
@@ -60,6 +60,7 @@ class Meta:
             "full_name",
             "gender",
             "profile_picture",
+            "email_address",
```
odinuge: This will potentially return the private email of users without @abakus.no-email. Don't you intend to use `internal_email_address` instead?

LudvigHz: Didn't realize the user email was private, will change.

PR #1455, opened 2019-02-20
[WIP] Add youtube_url field in article model

mrfarstad:
Needs more validation


lego/apps/articles/models.py
```diff
@@ -1,13 +1,18 @@
...
 from lego.apps.files.models import FileField
 from lego.apps.permissions.models import ObjectPermissionsModel
 from lego.utils.models import BasisModel
 
+from rest_framework.exceptions import ValidationError
+
 
 class Article(Content, BasisModel, ObjectPermissionsModel):
     cover = FileField(related_name="article_covers")
+    youtube_url = URLField(default="")
```
mrfarstad: You should add a custom validator here to check that the URL is for Youtube. See the `validators.py` file for inspiration

lego/apps/articles/migrations/0012_auto_20190304_1952.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 2.1.7 on 2019-03-04 19:52
```
orhanhenrik: You can merge all these migrations into one (0009 - 0012)

lego/utils/youtube_validator.py
```diff
@@ -0,0 +1,10 @@
+from django.core.validators import RegexValidator
+
+youtube_validator = RegexValidator(
```
orhanhenrik: Is this maybe a bit too strict? What if youtube changes their URLs a bit in the future?  
Is it not enough to just check that the domain is `youtube.com` or `youtu.be`?
Why are you using non-capturing groups? `(?:`

mrfarstad: I want the regex to match against the id of the youtube video as well, since we don't want video urls without ids 🤔

I use non-capturing groups, as I only want it to match the regex and not extract the result from the groups later. Not really sure if this is a good way to implement it, however. 😅

lego/apps/articles/migrations/0012_auto_20190304_1952.py
mrfarstad: How do I merge the migrations? 💯

lego/apps/articles/fixtures/development_articles.yaml
```diff
@@ -56,3 +53,4 @@
     tags:
       - lorem
       - ipsum
+    youtube_url: https://www.youtube.com/watch?v=bLHL75H_VEM
```
edvardvb: Do we need both articles in dev fixtures to have youtube headers? Could be annoying when developing other things. Maybe add a third article for youtube dev work, instead of changing the existing ones?

lego/apps/articles/serializers.py
```diff
@@ -33,6 +33,7 @@ class Meta:
             "content",
             "created_at",
             "pinned",
+            "youtube_url",
```
edvardvb: Correct me if I'm wrong, but this only adds the url when viewing a specific article right? What happens when a youtube article is on the frontpage? Or on the `/articles`-list view?

lego/apps/articles/tests/test_articles_api.py
```diff
@@ -24,6 +24,7 @@ def get_data_with_author(author_pk):
         "description": "good article",
         "content": "the best article",
         "author": author_pk,
+        "youtube_url": "https://www.youtube.com/watch?v=KrzIaRwAMvc",
```
edvardvb: As discussed, this can be removed. Both tests add a youtube url manually, and adding it here will affect all other tests using this method.

lego/apps/events/fixtures/development_events.yaml
```diff
@@ -4,39 +4,43 @@
...
-    text: "<p>TBA</p>"
+    text: '<p>TBA</p>'
     title: EINR
     is_priced: True
     price_member: 50
     price_guest: 100
     created_by: 3
     require_auth: False
     cover: test_event_cover.png
+    youtube_url: https://www.youtube.com/watch?v=bLHL75H_VEM
```
edvardvb: Same as for dev articles. You're changing the data people are using for developing other things as well, so it might be better to add a separate event for youtube-stuff. However, there are like 50 events to use , compared to only 2 articles, so it's not really a big deal here.

lego/apps/events/serializers/events.py
```diff
@@ -127,6 +127,7 @@ class Meta:
             "registration_count",
             "survey",
             "use_consent",
+            "youtube_url",
```
edvardvb: Same as articles, what happens on the front page or event-list when the backend doesn't pass a youtube url for the header?

lego/apps/joblistings/fixtures/development_joblistings.yaml
```diff
@@ -15,10 +15,11 @@
     visible_to: '2030-09-30T16:15:00+00:00'
     job_type: summer_job
     workplaces:
-    - 2
+      - 2
     from_year: 3
     to_year: 5
     application_url: 'https://bekk.no'
+    youtube_url: https://www.youtube.com/watch?v=bLHL75H_VEM
```
edvardvb: Same as article and events dev fixturs

lego/apps/events/serializers/events.py
edvardvb: Realized I might have misunderstood - is the idea to upload a header for frontpage/list use, and have the youtube video for the detail view?

lego/apps/articles/fixtures/development_articles.yaml
orhanhenrik: For sure!

lego/apps/events/serializers/events.py
mrfarstad: Yep, the idea was that the cover photo should be shown for lists and the youtube video will be shown on detail view 👍

lego/apps/articles/serializers.py
mrfarstad: Then the normal cover photo will be present 👍

lego/apps/articles/fixtures/development_articles.yaml
mrfarstad: Good idea! 👍

PR #1457, opened 2019-02-22
Combine undergrupper and interessegrupper into grupper

cherrybeans:
Since interestgroups were not really a substantial category, combine sub groups and interest groups into one category, groups.


orhanhenrik:
You need to make migrations. Also, you'll need to do this in multiple steps: first add the new category, then update all records that point to the old categories, then remove the old categories. You can either do this with multiple migrations and using [data migrations](https://docs.djangoproject.com/en/1.11/topics/migrations/#data-migrations) or just by making two PRs and updating the database in between.

cherrybeans:
Ahh, I thought I didn't have to since I just changed the choices, but it makes sense. I've already moved all of the articles that were pointed to the old ones manually, so that should be fine, right?

orhanhenrik:
How did you move the existing ones? The GROUPS category doesn't exist yet, does it?

docs/conf.py
```diff
@@ -20,181 +20,175 @@
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath('..'))
+sys.path.insert(0, os.path.abspath(".."))
```
orhanhenrik: why change this file?

cherrybeans: Seems like it was caught by black, should be fine?

cherrybeans:
> How did you move the existing ones? The GROUPS category doesn't exist yet, does it?

I moved them to Bedrifter temporarily

PR #1466, opened 2019-02-25
Fix frontpage

odinuge:
This would benefit of some tests :tm:


cherrybeans:
Is this the fix to why the frontpage is not working when not logged in?

odinuge:
Yes sir

cherrybeans:
What made it malfunction in the first place? Curious

odinuge:
https://github.com/webkom/lego/pull/1466/files#diff-20e67ca57d33bd498ade43a5f610580cR73

:arrow_up: @cherrybeans

cherrybeans:
Already saw that, not that smart though :disappointed:

odinuge:
It is because `Registration.objects.filter(user=request.user)` isn't valid when the user isn't authenticated.

PR #1475, opened 2019-03-04
Add logic for deleting comments

LudvigHz:


lego/apps/comments/models.py
```diff
@@ -27,5 +27,14 @@ class Meta:
         ordering = ("created_at",)
         permission_handler = CommentPermissionHandler()
 
+    def delete(self):
+        with transaction.atomic():
+            self.created_by = None
+            self.text = "<p>Slettet</p>"
```
odinuge: Does it make more sense to set this to `null`, and handle it in the frontend? Kinda hard to change the behavior if we save this string in the db.

(I have no idea if our `ContentField` will handle it properly without modification tho.)

lego/apps/comments/models.py
```diff
@@ -27,5 +27,14 @@ class Meta:
...
+            if Comment.objects.filter(parent=self).exists():
+                self.save()
+            else:
+                super(Comment, self).delete()
```
odinuge: Does it make more sense to do `delete(force=True)`. I don't see why this should be a `PersistentModel` at all

odinuge: Also, why set `created_by` and `text` before deleting (when using force)? Isn't it more logical to do that separatly? And, if soft deleting, does `.delete()` save the new values?

lego/apps/comments/models.py
```diff
@@ -27,5 +27,14 @@ class Meta:
         ordering = ("created_at",)
         permission_handler = CommentPermissionHandler()
 
+    def delete(self):
+        with transaction.atomic():
+            self.created_by = None
+            self.text = "<p>Slettet</p>"
+            if Comment.objects.filter(parent=self).exists():
```
odinuge: Nice :100:

lego/apps/comments/models.py
```diff
@@ -27,5 +27,14 @@ class Meta:
         ordering = ("created_at",)
         permission_handler = CommentPermissionHandler()
 
+    def delete(self):
+        with transaction.atomic():
```
orhanhenrik: Is there any reason to open a transaction here?

LudvigHz: I guess not really.

lego/apps/comments/models.py
```diff
@@ -1,6 +1,6 @@
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
-from django.db import models
+from django.db import models, transaction
```
odinuge: Is this import still used?

lego/apps/comments/models.py
```diff
@@ -15,7 +15,7 @@ def get_queryset(self):
 
 class Comment(BasisModel):
 
-    text = ContentField(allow_images=False)
+    text = ContentField(allow_images=False, null=True)
```
odinuge: Have you tested this (both frontend and backend)?

LudvigHz: Tested frontend, but can write a test for backend as well.

PR #1485, opened 2019-03-08
Change how event following is checked

JonasBak:
WIP, need to test that it actually works, need to fix when not logged in

### DIFF
--- a/lego/apps/followers/views.py
+++ b/lego/apps/followers/views.py
@@ -1,11 +1,13 @@
 from rest_framework import mixins, viewsets
 
+from lego.apps.events.models import Event
 from lego.apps.followers.models import FollowCompany, FollowEvent, FollowUser
 from lego.apps.followers.serializers import (
     FollowCompanySerializer,
     FollowEventSerializer,
     FollowUserSerializer,
 )
+from lego.apps.permissions.utils import get_permission_handler
 
 from .filters import FollowCompanyFilterSet, FollowEventFilterSet, FollowUserFilterSet
 
@@ -16,7 +18,10 @@ class FollowerBaseViewSet(
     mixins.DestroyModelMixin,
     viewsets.GenericViewSet,
 ):
-    pass
+    def get_queryset(self):
+        queryset = super().get_queryset()
+        permission_handler = get_permission_handler(queryset.model)
+        return permission_handler.filter_queryset(self.request.user, queryset)
 
 
 class FollowUserViewSet(FollowerBaseViewSet):
--- a/lego/apps/followers/filters.py
+++ b/lego/apps/followers/filters.py
@@ -1,4 +1,6 @@
+from django_filters import ModelChoiceFilter
 from django_filters.filterset import FilterSet
+from lego.apps.events.models import Event
 
 from .models import FollowCompany, FollowEvent, FollowUser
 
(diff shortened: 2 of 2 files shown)


lego/apps/followers/views.py
```diff
@@ -30,6 +31,13 @@ class FollowEventViewSet(FollowerBaseViewSet):
     queryset = FollowEvent.objects.all().select_related("follower")
     filter_class = FollowEventFilterSet
 
+    def get_queryset(self):
```
odinuge: Imo we should rethink how we _should_ do follow-stuff instead of just quick-filtering like this cases. There is a lot of stuff that should be fixed, and monkeypatching like this just makes the problem worse imo. :monkey: 

If there is a problem with this being time-sensitive, we can just disable it all together (like we did with feeds).

JonasBak: I agree that we maybe should rethink how we do following, but is this really that much of a monkeypatch when it's really just moving where the follower is filtered from the FilterSet to the View?

I also feel like disabling this won't be the same as disabling feeds, as this is a more used feature. Not that it is time sensitive, just annoying with all the errors.

odinuge: Still feels like monkeypatching of a generic api to me. :upside_down_face: The api is called `followers-event`, and this changes the behavior fundamentally. Since the following api is generic, `FollowUserViewSet`, `FollowEventViewSet` and `FollowCompanyViewSet` should work similar imo. :man_shrugging: 

I more or less agree that this is an ok solution, but it could also be handled using permissions - and we still have to write some permission code.

If this is just for fixing errors, we can use this solution: https://github.com/webkom/lego-webapp/pull/1480#discussion_r263528879. We can also just ignore them, if that is the problem.

JonasBak: Is there any reason not to put this in the `FollowerBaseViewSet` so that the three views work the same way? I agree that this changes what the api is, but considering how these views are actually used I don't see any particular reasons why the view should be so generic that you are able to filter on any user anyways.

If we are going to do more with following in the future I think the system probably need a larger rework anyways, also introducing permissions, but as far as I know we don't have any plans of doing that (especially now that feeds are disabled) in some time, and I feel like we keep things generic just for the sake of it being generic.

odinuge: Putting the logic in `FollowerBaseViewSet` is way better. I still think this kind of logic should be handled with permissions, and not with simple qs-filtering here. A few lines in the permission handler can do the same thing as this, in a more elegant way.

I don't remember what we agreed upon about following users & events tho.

However, the most important thing: `following` needs tests!!

JonasBak: Yeah, I agree that it's better to put the logic in the base view, I just wanted to check that it worked the way I thought, so I made a quick wip pr. And I agree that permissions is the way to go. I can write some test as well :+1:

lego/apps/followers/views.py
```diff
@@ -16,7 +18,10 @@ class FollowerBaseViewSet(
     mixins.DestroyModelMixin,
     viewsets.GenericViewSet,
 ):
-    pass
+    def get_queryset(self):
+        queryset = super().get_queryset()
+        permission_handler = get_permission_handler(queryset.model)
```
odinuge: This can also be archived implicitly by updating the permission handler to do this automagically with permissions.

odinuge:
Guess we can merge this now in order to fix the annoying errors in the webapp. This will essentially disable keyword-perms, and we should reflect that in the permission handler instead of just filtering the queryset manually (this applies to quite a few other apps too). This will indeed work, so it should be good to go.

PR #1488, opened 2019-03-11
Add event_status_type for events

### LINKED ISSUE #1378: "Registration open" when it's not
All events in https://abakus.no/events are labeled as "Påmelding åpen!", even when they're more than a week away

screenshot 2019-01-30 14 57 55

kotlarz:
Fixes #1378.

- [x] Add event status type: `TBA`.
Events marked as `TBA` should have 0 pools, and set the location to `TBA`.
A user should **not** be able to sign up to this event.
There are no permissions (except Abakom).
This event status type should be used for events that are not completed yet (information to be added), this should also be the default event status type.
**[tests added]**
- [x] Add event status type: `OPEN`.
Events marked as `OPEN` should have 0 pools, just like `TBA`. The only difference is that `OPEN` allows a location to be set.
A user should **not** be able to sign up to this event.
There are no permissions (except Abakom).
This event status type should be used for events hosted by LaBamba, etc.
**[tests added]**
- [x] Add event status type: `INFINITY`.
Events marked as `INFINITY` should have 1 pool, with `capacity` set to `0`.
A user **should** be able to sign up to this event.
This event status type should be used for events such as Abakom Works, etc.
**[tests added]**
- [x] Add event status type: `NORMAL`.
Events marked as `NORMAL` should have function as events previously worked before this PR.
**[tests exists]**


lego/apps/events/constants.py
```diff
@@ -18,6 +18,12 @@
     (EVENT, EVENT),
 )
 
+NORMAL = "normal"
+INFINITY = "infinity"
```
JonasBak: Feel like `INFINITE` fits better, as it is an "infinite pool" not "infinity pool"

lego/apps/events/constants.py
```diff
@@ -18,6 +18,12 @@
...
+OPEN = "open"
+TBA = "tba"
+EVENT_STATUS_TYPES = ((NORMAL, NORMAL), (INFINITY, INFINITY), (OPEN, OPEN), (TBA, TBA))
```
JonasBak: Can we get some documentation somewhere that explains what the different statuses represent? I feel like it's not that self explanatory, and it would be nice to not have to look up this pr to read what they mean.

lego/apps/events/serializers/events.py
```diff
@@ -269,6 +281,15 @@ def validate(self, data):
     def create(self, validated_data):
         pools = validated_data.pop("pools", [])
         is_abakom_only = validated_data.pop("is_abakom_only", False)
+        event_status_type = validated_data.get("event_status_type", constants.NORMAL)
+        if event_status_type == constants.TBA:
+            pools = []
+            validated_data["location"] = "TBA"
```
JonasBak: Shouldn't you be able to set a location if you have one, even if the event itself is TBA?

SmithPeder: I don't think so. This will open more cases. TBA events that have a location, and TBA events that don't have a location. I think we should make it clear and simple, this is an event that is to be announced.

kotlarz: I honestly agree, it's easier for us just to set the location as TBA, and let them rather set the location once it's no longer TBA.

I'll update the PR to reflect this tomorrow.

SmithPeder: 90% of all TBA events have 'TBA' as location anyhow.

JonasBak: Isn't the point of this to not have to use the location to determine what type of event, so is it really another case to have a location?

lego/apps/events/migrations/0024_auto_20190311_2307.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 2.1.7 on 2019-03-11 23:07
```
orhanhenrik: You can merge these migrations into one

lego/apps/events/serializers/events.py
orhanhenrik: I don't see a reason to restrict the location on events of type TBA. What function does it have? We can style the TBA event in frontend, and you cannot create pools unless the event is not changed to normal mode anyways.

lego/apps/events/tests/test_events_api.py
```diff
@@ -419,6 +481,76 @@ def test_event_creation(self):
             for key in ["name", "capacity", "activationDate", "permissionGroups"]:
                 self.assertEqual(res_pools[i][key], expect_pools[i][key])
 
+    def test_event_creation_tba(self):
+        """Test event creation for TBA status type"""
+        self.event_response = self.client.post(_get_list_url(), _test_event_data[2])
+        self.assertEqual(self.event_response.status_code, 201)
+        self.event_id = self.event_response.data.pop("id", None)
```
orhanhenrik: Why use `self. event_id` instead of just `event_id`?

lego/apps/events/tests/test_events_api.py
```diff
@@ -419,6 +481,76 @@ def test_event_creation(self):
             for key in ["name", "capacity", "activationDate", "permissionGroups"]:
                 self.assertEqual(res_pools[i][key], expect_pools[i][key])
 
+    def test_event_creation_tba(self):
+        """Test event creation for TBA status type"""
+        self.event_response = self.client.post(_get_list_url(), _test_event_data[2])
+        self.assertEqual(self.event_response.status_code, 201)
```
orhanhenrik: Shouldn't we return 400 if the user tries to set location and pools instead of "failing" silently?

lego/apps/events/tests/test_events_api.py
```diff
@@ -419,6 +481,76 @@ def test_event_creation(self):
...
+            "location",
+            "end_time",
+            "merge_time",
+            "is_abakom_only",
+        ]:
+            self.assertEqual(res_event[key], expect_event[key])
+        created_event = Event.objects.get(id=self.event_id)
+        self.assertEqual(created_event.event_status_type, "infinite")
+        self.assertEqual(len(created_event.pools.all()), 1)
+        self.assertEqual(created_event.pools.all().first().capacity, 0)
```
orhanhenrik: You can omit `.all()` when calling `first()`

lego/apps/events/tests/test_events_api.py
```diff
@@ -419,6 +481,76 @@ def test_event_creation(self):
...
+            "end_time",
+            "merge_time",
+            "is_abakom_only",
+        ]:
+            self.assertEqual(res_event[key], expect_event[key])
+        created_event = Event.objects.get(id=self.event_id)
+        self.assertEqual(created_event.event_status_type, "open")
+        self.assertEqual(len(created_event.pools.all()), 0)
+
+    def test_event_creation_infinite(self):
```
orhanhenrik: Should also add a test case that checks that `normal` type is used as default if no `event_status_type` is passed

lego/apps/events/serializers/events.py
SmithPeder: It's just less to fill out when creating a `TBA` event. Now they need to look up if they maybe have a location and fill that since it's a required field or write TBA themselves.

And when switching to a normal event I want the `location` field to be empty, so they don't forget to switch that from `TBA/random`.

lego/apps/events/tests/test_events_api.py
SmithPeder: Created a test: line 573

lego/apps/events/tests/test_events_api.py
SmithPeder: I think the backend should just overwrite the input and set the event to be TBA with `pools = []` and return `201`. That way frontend bugs where location and events suddenly retain their value or something won't crash the application.

lego/apps/events/migrations/0023_auto_20190319_0007.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 2.1.7 on 2019-03-19 00:07
```
orhanhenrik: merge these migrations

lego/apps/events/tests/test_events_api.py
```diff
@@ -419,6 +500,83 @@ def test_event_creation(self):
...
+            "is_abakom_only",
+        ]:
+            self.assertEqual(res_event[key], expect_event[key])
+        created_event = Event.objects.get(id=self.event_id)
+        self.assertEqual(created_event.event_status_type, "INFINITE")
+        self.assertEqual(len(created_event.pools.all()), 1)
+        self.assertEqual(created_event.pools.first().capacity, 0)
+
+    def test_event_create_no_event_status_type(self):
+        """Test evnet creation with no event status type posted"""
```
orhanhenrik: evnet

lego/apps/events/models.py
```diff
@@ -43,6 +43,9 @@ class Event(Content, BasisModel, ObjectPermissionsModel):
     """
 
     event_type = models.CharField(max_length=50, choices=constants.EVENT_TYPES)
+    event_status_type = models.CharField(
+        max_length=50, choices=constants.EVENT_STATUS_TYPES, default=constants.TBA
```
orhanhenrik: Why is the default here TBA while the default in the api endpoint is NORMAL?

lego/apps/events/migrations/0023_auto_20190319_0007.py
odinuge: `TBA` should be default for new event, while old events should be migrated ti get `NORMAL`.

lego/apps/events/serializers/events.py
```diff
@@ -281,6 +302,15 @@ def create(self, validated_data):
...
         pools = validated_data.pop("pools", None)
         is_abakom_only = validated_data.pop("is_abakom_only", False)
+        event_status_type = validated_data.get("event_status_type", constants.NORMAL)
+        if event_status_type == constants.TBA:
+            pools = []
+            validated_data["location"] = "TBA"
+        elif event_status_type == constants.OPEN:
+            pools = []
+        elif event_status_type == constants.INFINITE:
+            for pool in pools:
```
odinuge: Maybe verify that this case only has one pool?

PR #1497, opened 2019-03-14
Fix poll permissions

LudvigHz:
Fix poll permissions so that they actually are correct:
anonymous user (not logged in): no permissions
All authenticated users (logged in): list, view and vote
Admin users with correct permissions: update and create

Tests are also updated to match this. (and it doesn't force authenticate all users..)

This fixes issue where users get an error when trying to access abakus.no/polls/{id}

### DIFF
--- a/lego/apps/polls/tests/test_api_polls.py
+++ b/lego/apps/polls/tests/test_api_polls.py
@@ -23,11 +23,11 @@ class PollViewSetTestCase(BaseAPITestCase):
 
     def setUp(self):
         """Create test users"""
-        self.authenticated_user = User.objects.get(username="test1")
+        self.admin = User.objects.get(username="test1")
         self.group = AbakusGroup.objects_with_text.get(name="PollAdminTest")
-        self.group.add_user(self.authenticated_user)
+        self.group.add_user(self.admin)
 
-        self.unauthenticated_user = User.objects.get(username="test2")
+        self.pleb = User.objects.get(username="test2")
 
         self.poll_data = {
             "title": "Hva mener du?",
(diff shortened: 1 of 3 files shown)


lego/apps/polls/tests/test_api_polls.py
```diff
@@ -46,55 +46,53 @@ def setUp(self):
 
     def test_create_poll_authenticated(self):
         """A user with permissions should be able to create a poll"""
-        self.client.force_authenticate(self.authenticated_user)
+        self.client.force_login(self.authenticated_user)
```
odinuge: Is there a reason to change this to `force_login`? The "django-rest-framework" way of doing it is `force_authenticated`, while `force_login` is from the base django test case.

lego/apps/polls/tests/test_api_polls.py
```diff
@@ -113,17 +111,16 @@ def test_vote_poll_authenticated_answered(self):
...
         response = self.client.post(_get_vote_url(1), {"optionId": 1})
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(Poll.objects.get(pk=1).total_votes, votes + 1)
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+        self.assertEqual(Poll.objects.get(pk=1).total_votes, votes)
 
     def test_vote_poll_not_valid(self):
         """A user should not be able to vote on a poll where valid_until is not in the future"""
-        self.client.force_authenticate(self.authenticated_user)
+        self.client.force_login(self.unauthenticated_user)
```
odinuge: This user should be renamed to `normal_user` or `pleb` (we mostly use pleb quite a bit). `unauthenticated_user` implies that there is no user (aka. not authenticated - Anonymous user https://docs.djangoproject.com/en/2.1/ref/contrib/auth/#anonymoususer-object).

Also, `self.authenticated_user` should be renamed to something like `admin`.

Maor info about authentication vs authorization: https://stackoverflow.com/questions/6556522/authentication-versus-authorization

PR #1509, opened 2019-03-18
Bump stripe version to support Google & Apple pay

odinuge:
Screenshot from 2019-03-18 21-48-10

### DIFF
--- a/lego/apps/events/tests/test_payments.py
+++ b/lego/apps/events/tests/test_payments.py
@@ -20,8 +20,8 @@ class StripePaymentTestCase(BaseAPITestCase):
     fixtures = [
         "test_abakus_groups.yaml",
         "test_companies.yaml",
-        "test_events.yaml",
         "test_users.yaml",
+        "test_events.yaml",
     ]
 
     def setUp(self):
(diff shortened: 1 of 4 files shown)


lego/apps/events/tests/test_payments.py
```diff
@@ -20,8 +20,8 @@ class StripePaymentTestCase(BaseAPITestCase):
     fixtures = [
         "test_abakus_groups.yaml",
         "test_companies.yaml",
-        "test_events.yaml",
         "test_users.yaml",
+        "test_events.yaml",
```
edvardvb: What's up here? Was the previous ordering causing issues when loading fixtures for the tests?

odinuge: Yup. `2017/11/13 21:52:34` was the last time we executed the stripe tests :tm:

edvardvb: It worked on release, so why would we ever test it again??? 🙃

odinuge: Hahaha. Stripe has a quite stable API, so it shouldn't be a huge problem :laughing:

PR #1551, opened 2019-04-02
Stop living on the edge

orhanhenrik:
Old drones work better than new drones. New drones are too smart and cause more trouble. Let's try updating our robots again later.


edvardvb:
MADA 🇺🇸🍊

PR #1553, opened 2019-04-03
Add /previous/ endpoint.

HoboKristian:
Used to fetch previous events which a user was registered for.

This is needed to support a quicker backend compared to loading all pages and then filtering on the frontend.

### DIFF
--- a/lego/apps/events/views.py
+++ b/lego/apps/events/views.py
@@ -198,6 +198,39 @@ class EventViewSet(AllowedPermissionsMixin, viewsets.ModelViewSet):
         serializer_class=EventUserRegSerializer,
         permission_classes=[permissions.IsAuthenticated],
     )
+    def previous(self, request):
+        queryset = (
+            self.get_queryset()
+                .filter(
+                registrations__status=constants.SUCCESS_REGISTER,
+                registrations__user=request.user,
+                start_time__lt=timezone.now(),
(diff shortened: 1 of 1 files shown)


lego/apps/events/views.py
```diff
@@ -193,6 +193,39 @@ def payment(self, request, *args, **kwargs):
...
+    @decorators.action(
+        detail=False,
+        serializer_class=EventUserRegSerializer,
+        permission_classes=[permissions.IsAuthenticated],
+    )
+    def previous(self, request):
+        queryset = (
+            self.get_queryset()
+            .filter(
+                registrations__status=constants.SUCCESS_REGISTER,
```
edvardvb: Is this filtering sufficient? Won't this return events where the user is in the waiting list as well? @odinuge

odinuge: Correcto.

This is the same as `upcoming`, and imo. it's nice to keep them similar. (@HoboKristian only dispalys  events with pool here: https://github.com/webkom/lego-webapp/pull/1541/files#diff-ac325e8453fd68444445e7f1267ead7cR372)

We can make a separate viewset, or add some custom filtering to `/api/v1/events/` to make it more generic.

edvardvb: Ah, I see! 👍

PR #1555, opened 2019-04-08
[Do not merge] Update to python 3.7 (tox and prod)

edvardvb:
Updates the prod Dockerfile and tox to use Python 3.7.

However, the lego-testbase Docker image(https://github.com/webkom/lego-testbase) is based on Python 3.6, so this should not be merged without updating the test image. Works locally though!

Also this must be tested rigorously in staging to ensure everything works well with Python 3.7. Especially Celery. Perhaps we need some more methodic testing as well?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6
+FROM python:3.7
 MAINTAINER Abakus Webkom <webkom@abakus.no>
 
 ARG RELEASE
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,7 @@ envlist = tests, docs, missing-migrations, isort, flake8, coverage, black
 skipsdist = True
 
 [testenv]
-basepython = python3.6
+basepython = python3.7
 deps =
     black: -r{toxinidir}/requirements/black.txt
     isort: -r{toxinidir}/requirements/isort.txt
@@ -49,7 +49,7 @@ commands =
     python manage.py missing_migrations
 
 [testenv:coverage]
-basepython = python3.6
+basepython = python3.7
 deps = -r{toxinidir}/requirements/coverage.txt
 commands =
     coverage report --fail-under=75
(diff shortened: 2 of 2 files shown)


SmithPeder:
I bumped the `lego-testbase` image to python3.7 with a new tag so it doesn't mess with `:latest`. It seems to be working good. We should still make the 3.7 latest before this is merged, then revert my commit here. 

I wanna get the `python-alpine` version to work tho. I think all it's missing is the `make` function, and in 1/10th of the size 😱 
image

odinuge:
:+1:

3.7 is the way to go! We should use the same version for both tests and when running in prod. We still run 3.6 https://github.com/webkom/lego/blob/a7b26d93a80992803ebf4bc0eafd5e1932e586da/Dockerfile#L1

Before switching the dockerfile to 3.7 we should do some battle testing to verify that stuff works (i have had some strange issues with celery and python 3.7 when running without eager, but I guess it should work fine).

odinuge:
Same for alpine @SmithPeder. We should use the same base img (debian, alpine, ubuntu etc.) for the tests and the final build. We had some strange dns issued with alpine last time we tried it. ( I <3 alpine tho.)

SmithPeder:
Using the same image for testing and final build sound like a good philosophy. I tried building the lego image on 3.7 base and it seems to be working here. But as you said, it should be tested in the full environment with the other images.

SmithPeder:
@edvardvb  Maybe append https://github.com/webkom/lego/pull/1580 as well?

edvardvb:
Hm. It's possible, but not really necessary. This PR focuses on moving us to python 3.7, and it might be better to test that separately from other changes. If we run into issues with tox3.9.0 we won't know if python 3.7 or tox is the issue!

SmithPeder:
Mr.Master over here with his [scientific methods](https://en.wikipedia.org/wiki/Scientific_method).

> It involves formulating hypotheses, via induction, based on such observations; experimental and measurement-based testing of deductions drawn from the hypotheses; and refinement (or elimination) of the hypotheses based on the experimental findings.

odinuge:
lgtm

PR #1571, opened 2019-04-09
Update black to 19.3b0

edvardvb:


edvardvb:
`make fixme` still works

PR #1593, opened 2019-04-29
Set lego-testbase image to p37 temp

SmithPeder:


SmithPeder:
nwm

odinuge:
Drone will cache the images. Use this instead of pull: always

PR #1608, opened 2019-05-10
Add/created by and updated by for event registrations

kotlarz:


odinuge:
I did however read one more line in the docs today, compared to yesterday, and it looks like there is a `defaults`-kwarg in `get_or_create` for args that are only for `create`, and _not_ get... :laughing: Just like we wanted yesterday :upside_down_face:

kotlarz:
> I did however read one more line in the docs today, compared to yesterday, and it looks like there is a `defaults`-kwarg in `get_or_create` for args that are only for `create`, and _not_ get...  Just like we wanted yesterday

Ooooh, yeah, that would have saved me a bit of bother yesterday heh.

> Any keyword arguments passed to get_or_create() — except an optional one called defaults — will be used in a get() call.

https://docs.djangoproject.com/en/2.1/ref/models/querysets/

PR #1643, opened 2019-08-19
Set request scheme for cors in dev

JonasBak:
So I actually can migrate and start backend locally


kotlarz:
Oops :upside_down_face:

Issue #1508, opened 2019-03-18
Add proper scope support for oauth!


odinuge:
Add proper support for scopes in oauth. Current impl. will give full access, regardless of what the scope is provided.

odinuge:
Imo this is ~blocking for _committee-admissions_, so it should be fixed before opening for new admissions.

odinuge:
To begin with, I think we can deny all requests from oauth2 tokens (without the (new) `all` scope), and create a separate endpoint with basic user info that oauth2 apps can use:
- Name
- Username
- pk
- profile picture
- Email (do we need that??)
- memberships w/nested groups
- isActive
- isStudent
- isAbakusMember
- isAbkomMember

cc @cherrybeans

odinuge:
Initial implementation with two scopes fixed in #1645 

Scopes:
- `all`: Do all the things
- `user`: Read access to baisc user info (`/api/v1/users/oauth2_userdata`).

PR #1645, opened 2019-08-22
Add basic scope support to oauth

odinuge:
If the an oauth app use scope "user", it will only be granted access to
"/users/oauth2_userdata", a simple endpoint with name, profile picture,
memberships.

Scope "all" gives full access.


odinuge:
This fixes #1508
Screenshot from 2019-08-22 20-59-15

Issue #1647, opened 2019-08-24
Add reactions to Articles


kotlarz:

kotlarz:
Closed by https://github.com/webkom/lego-webapp/pull/1589 and https://github.com/webkom/lego/pull/1659

Issue #1448, opened 2016-09-30
<Reactions />


hanse:
Make a React component `<Reactions />` that show a popup with clickable reactions when the trigger element is clicked.

PR #1659, opened 2019-09-01
Implement a proper backend for Emojis and Reactions

kotlarz:
This implements the backend for https://github.com/webkom/lego/issues/1448 and https://github.com/webkom/lego/issues/1647.

The following things have been done/implemented:
- Emojis have been split out from Reactions (ReactionType has been renamed to Emoji)
- Emojis fixtures have been replace with [emojilib](https://github.com/muan/emojilib), instead of [EmojiOne](https://github.com/joypixels/emojione) (deprecated).
- Emojis now have additional fields (keywords, category, i.e.)
- There is a limit on how many reactions an user can react with per object (i.e. an Article), the max is currently 10, see `lego/apps/reactions/constants.py`.
- The users that have reacted to something have been hidden, we only need the count/amount of reactions.
- All **1570** emojis are enabled by default, but they can be disabled (there's `disabled` field on the Emoji model).

The API routes are now:
- `HTTP GET /api/v1/emojis` to list all of the Emojis (not paginated).
- `HTTP POST /api/v1/reactions` to create a new Reaction.
- `HTTP DELETE /api/v1/reactions/<reaction_id>` to delete a Reaction.

Further work:
- Implement WebSocket to allow live updates on the frontend.
- Implement custom Emojis.
We could implement it in the way we do it with Quotes, where admins can approve/unapprove emojis.
Custom emojis have to be in an image format, i.e. 32px * 32px transparent png.


lego/apps/content/models.py
```diff
@@ -78,3 +84,7 @@ def __str__(self):
     @property
     def comment_target(self):
         return f"{self._meta.app_label}.{self._meta.model_name}-{self.pk}"
+
+    @property
+    def reaction_target(self):
```
orhanhenrik: Should we combine comment_target and reaction_target? I can't imagine a case where they would be different.

lego/apps/emojis/models.py
```diff
@@ -0,0 +1,21 @@
...
+from django.db import IntegrityError, models
+from rest_framework.exceptions import ValidationError
+
+from lego.apps.emojis.permissions import EmojiPermissionHandler
+from lego.utils.managers import BasisModelManager
+from lego.utils.models import BasisModel
+
+
+class Emoji(BasisModel):
+    id = models.CharField(max_length=40, primary_key=True)
```
orhanhenrik: Maybe this should be named something more descriptive? I wouldn't know from the name `id` that this field contains e.g. `":smile:"`

lego/apps/emojis/models.py
```diff
@@ -0,0 +1,21 @@
...
+    keywords = fields.ArrayField(models.CharField(max_length=40))
+    unicode_string = models.CharField(max_length=24, db_index=True)
+    fitzpatrick_scale = models.BooleanField()
```
orhanhenrik: How do we handle reactions with emojis that can have different skin colors?

lego/apps/emojis/views.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework import mixins, viewsets
+
+from lego.apps.emojis.models import Emoji
+from lego.apps.emojis.permissions import EmojiPermissionHandler
+from lego.apps.emojis.serializers import EmojiSerializer
+from lego.apps.permissions.api.views import AllowedPermissionsMixin
+
+
+class EmojiViewSet(AllowedPermissionsMixin, viewsets.ModelViewSet):
+    pagination_class = None
```
orhanhenrik: Is there supposed to be pagination here?
From PR description:
> -  HTTP GET /api/v1/emojis to list all of the Emojis (paginated).

lego/apps/reactions/exceptions.py
```diff
@@ -0,0 +1,20 @@
...
+from rest_framework.exceptions import APIException
+
+
+class APIReactionExists(APIException):
+    status_code = status.HTTP_409_CONFLICT
+    default_detail = "Reaction already exists."
+
+
+class APITooManyReactions(APIException):
+    status_code = status.HTTP_413_REQUEST_ENTITY_TOO_LARGE
```
orhanhenrik: This status code is maybe a bit weird? 🤔

lego/apps/reactions/serializers.py
```diff
@@ -1,31 +1,43 @@
...
 from lego.utils.serializers import BasisModelSerializer, GenericRelationField
 
 
-class ReactionTypeSerializer(BasisModelSerializer):
-    class Meta:
-        model = ReactionType
-        fields = ("short_code", "unicode")
-
-
 class GroupedReactionSerializer(serializers.Serializer):
```
orhanhenrik: Is this serializer used?

lego/apps/content/models.py
```diff
@@ -55,19 +55,25 @@ class Content(SlugModel):
 
     slug_field = "title"
 
-    @property
-    def reactions_grouped(self):
+    def get_reactions_grouped(self, user):
```
odinuge: I suspect this will be kinda non-optimal, as it will result in a few more queries than needed. Not so important now, but I think most of the logic can be done in sql instead of python

lego/apps/content/models.py
kotlarz: Not a bad idea, I'll rename it to `content_target`

lego/apps/emojis/models.py
kotlarz: I agree somewhat, in that case we just need a descriptive name, perhaps `short_code` as it was previous named. I would however like some help with handling the `byId` in the frontend, as I'm not sure how to handle index it with other fields than `id`.

lego/apps/emojis/models.py
kotlarz: That's what this field is for, however it isn't used now. I'd rather focus on it later on, might require some unique constrains if we decide on using skin colors.

lego/apps/emojis/models.py
odinuge: Like this (using tag as the pk in normalizer)?
https://github.com/webkom/lego-webapp/blob/master/app/reducers/index.js#L219

lego/apps/emojis/views.py
kotlarz: Brain broke, I mean to write that there's no pagination.

lego/apps/reactions/exceptions.py
kotlarz: I went through a few status codes to look for a proper code for it. What would you suggest? It's for when a user has reacted too many times in a content object. I considered using 409 Conflict instead, but I don't believe that's fitting either.

lego/apps/emojis/fixtures/emojis.json
```diff
@@ -0,0 +1,9422 @@
+{
```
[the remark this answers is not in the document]
kotlarz: Will do 👍

lego/apps/content/models.py
kotlarz: Agreed, I just went built on what was previously created. I will consider moving the logic if I have the time, or fix it in a future PR.

lego/apps/reactions/models.py
```diff
@@ -34,3 +33,26 @@ def __repr__(self):
...
+                created_by=self.created_by,
+            ).count()
+            >= REACTION_COUNT_LIMIT
+        ):
+            raise TooManyReactions()
+        else:
+            super(Reaction, self).save(*args, **kwargs)
+
+    def delete(self, using=None, force=False):
+        super().delete(using, force=True)
```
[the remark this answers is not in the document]
kotlarz: Only issue is that I want the `created_by` field from BasisModel, which extends the PersistentModel. I could remove the extension and just add the `created_by` to the Reaction model. What do you think?

lego/apps/reactions/serializers.py
kotlarz: It is not currently, the plan was to use it once the logic to get the grouped reactions has been moved to sql. I'll remove the serializer now.

lego/apps/reactions/models.py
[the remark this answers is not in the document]
odinuge: Yee, I know (even tho I don't understand why). Keep it like this for now, and then we can (maybe :trollface:) look into it some time later. This applies to ~3-4 things from what I can remember.

lego/apps/reactions/exceptions.py
kotlarz: I do agree 413 isnt really fitting for a real REST api, so I'm open to suggestions.

lego/apps/reactions/models.py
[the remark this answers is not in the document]
kotlarz: Might be worth creating a new model for it then :man_shrugging:
#todo

lego/apps/reactions/exceptions.py
odinuge: `409` is reasonable, but I don't really care that much about return codes (as they are mostly stupid). As long as the message is ok, and in this case it is, it doesn't really matter.

As long as forms utilize the form validation data in the cases it makes sense, everything should be ok.

kotlarz: I'm keeping it as `413` for now, as I use it for error handling for the frontend.

lego/apps/content/models.py
orhanhenrik: Hm, we also need to keep in mind that renaming it is not backwards compatible, so the backend needs to be deployed together with the frontend. If one is deployed before the other, comments will break.

lego/utils/renderers.py
```diff
@@ -8,7 +8,7 @@ class JSONRenderer(CamelCaseJSONRenderer):
     """
 
     def render(self, data, *args, **kwargs):
-        if data is None:
+        if data is None and len(args) > 1 and args[1]["response"].status_code != 204:
```
odinuge: :heart_eyes:

kotlarz: best code

Issue #1554, opened 2019-04-04
Set status payment status on old registrations (from NERD) to "manual"


odinuge:
Only valid registrations with pool on paid events should be marked "manual".

Just like: https://abakus.no/events/1767

![Screenshot from 2019-04-04 15-15-47](https://user-images.githubusercontent.com/1467188/55558519-8ed36500-56ec-11e9-964a-585c6e07e9c4.png)

ivarnakken:
Please enlighten me, but is this still needed now? 😄

LudvigHz:
> Please enlighten me, but is this still needed now? smile @odinuge
> 
> @LudvigHz, you might know?

The reason why this issue is here is that in the event we change some code or decide to check unpaid registrations, all these old things would be included and people could get emails and notifications of unpaid events from ~10 years ago. 

As it is now, it hasn't been an issue. But that's not really a strong argument, so we might still want to look into this. It shouldn't be too hard either. It's just about creating a migration and setting the payment status on all these old registrations.

TLDR; we probably _should_ fix this, but if it's critical? :shrug:

PR #1669, opened 2019-09-10
Switch to Stripe payment intent API to support SCA

LudvigHz:
Must be merged at the same time as https://github.com/webkom/lego-webapp/pull/1590

#### TODO
- [x] Edit tests
- [x] Handle unregistration and canceling the payment intent


kotlarz:
How do I test this locally?

LudvigHz:
> How do I test this locally?

To test this you need to start the server with `STRIPE_TEST_KEY={our stripe test key}`, which is on https://dashboard.stripe.com/test/apikeys

In order to test with webhooks, you need to do the following
- In `lego/apps/events/webhooks.py`
  ```py
  from django.views.decorators.csrf import csrf_exempt
  ...
   # in StripeWebhookPermission
   # ignore the signature verification as you need the secret
       def has_permission(self, request, view):
          ...
          if not (secret and request_header):
               return True # add this line
               ...

   # add the decorator to the webhook view
   @csrf_exempt
   class StripeWebhook(...):
   ...
  ```
- Use a http forwarder and add the forwarder url with the endpoint `api/v1/webhooks-stripe/` to the stripe dashboard in test mode: https://dashboard.stripe.com/test/webhooks

lego/apps/events/tasks.py
```diff
@@ -184,11 +190,7 @@ def registration_payment_save(self, result, registration_id, logger_context=None
         registration.charge_amount = result["amount"]
         registration.charge_status = result["status"]
         registration.save()
-        notify_user_payment(
-            constants.SOCKET_PAYMENT_SUCCESS,
-            registration,
-            success_message="Betaling gjennomført",
-        )
+        return {"client_secret": result["client_secret"]}
```
JonasBak: Is this safe to pass around with celery?
https://stripe.com/docs/api/payment_intents/object#payment_intent_object-client_secret
> The client secret can be used to complete a payment from your frontend. _It should not be stored_, ...

lego/apps/events/constants.py
```diff
@@ -83,7 +83,7 @@
 
 PAYMENT_PENDING = "pending"
 PAYMENT_SUCCESS = "succeeded"
-PAYMENT_FAILURE = "failed"  # Matches stripe API
+PAYMENT_FAILURE = "payment_failed"  # Matches stripe API
```
JonasBak: Do we have to migrate old payments if they are stored as failed payments?

lego/apps/events/tasks.py
LudvigHz: I think it should be ok, it's not logged anywhere, and the only thing you can do with it is pretty much confirm a payment.

lego/apps/events/constants.py
LudvigHz: I realized this is actually wrong: https://stripe.com/docs/payments/intents#intent-statuses
I'm not sure how to handle this, since the payment intent does not have a `failed` status. We could set the status to failed if the charge object has a `failed` status, since we dont use the same payment intent to retry payments. Not the best solution though.
Or we could try to detect when a payment intent has been tried once, and manually cancel it?
Thoughts?

odinuge: Hmm.

I haven't look that much at Payment intents, but in the link :arrow_up: it looks like the flow is quite different from the current one. If that _is_ the case, I think we should rethink the payment process instead of just fitting it into the previous model.

Also, just like @JonasBak says, we would have to migrate the data no matter what. We also need to fix https://github.com/webkom/lego/issues/1554 somehow. It will be bad S if we suddenly send out thousands of mails saying that people have to pay for stuff they have already paid for.

lego/apps/events/serializers/registrations.py
```diff
@@ -157,6 +157,6 @@ class StripeMetaSerializer(serializers.Serializer):
 class StripeObjectSerializer(serializers.Serializer):
     id = serializers.CharField()
     amount = serializers.IntegerField()
-    amount_refunded = serializers.IntegerField()
+    amount_refunded = serializers.IntegerField(required=False)
```
odinuge: Add some refs to the docs here too.

lego/apps/events/tasks.py
```diff
@@ -226,9 +236,19 @@ def stripe_webhook_event(self, event_id, event_type, logger_context=None):
...
         registration.charge_amount = serializer.data["amount"]
-        registration.charge_amount_refunded = serializer.data["amount_refunded"]
+        registration.charge_amount_refunded = (
+            serializer.data["amount_refunded"]
+            if "amount_refunded" in serializer.data
+            else 0
+        )
         registration.charge_status = serializer.data["status"]
         registration.save()
+        if event_type == "payment_intent.succeeded":
```
odinuge: Hmm :thinking:

lego/apps/events/tasks.py
```diff
@@ -226,9 +236,19 @@ def stripe_webhook_event(self, event_id, event_type, logger_context=None):
             raise WebhookDidNotFindRegistration(event_id, metadata)
         registration.charge_id = serializer.data["id"]
         registration.charge_amount = serializer.data["amount"]
-        registration.charge_amount_refunded = serializer.data["amount_refunded"]
+        registration.charge_amount_refunded = (
```
odinuge: Hmm :thinking:

lego/apps/events/views.py
```diff
@@ -182,16 +179,27 @@ def payment(self, request, *args, **kwargs):
...
         if registration.has_paid():
             raise APIPaymentExists()
+
         registration.charge_status = constants.PAYMENT_PENDING
         registration.save()
-        chain(
-            async_payment.s(registration.id, serializer.data["token"]),
-            registration_payment_save.s(registration.id),
-        ).delay()
+        client_secret = (
```
odinuge: This makes ~no sense, as it will block the main thread to the celery worker completes, and we should _not_ do that. If that is the intention, it makes no sense to do it in celery instead of here directly. (ref. comment about rethinking the process)

odinuge: Also I don't understand the process here, as the mappings are related to the previous process.

lego/apps/events/views.py
```diff
@@ -182,16 +179,27 @@ def payment(self, request, *args, **kwargs):
...
+            .delay()
+            .get()
+        )
         payment_serializer = RegistrationPaymentReadSerializer(
             registration, context={"request": request}
         )
-        return Response(data=payment_serializer.data, status=status.HTTP_202_ACCEPTED)
+
+        response_data = payment_serializer.data
+        if client_secret:
```
odinuge: What about the error handling, when `!client_secret`?

lego/apps/events/tasks.py
odinuge: This feels bad

lego/apps/events/constants.py
LudvigHz: The flow actually isn't __that__ much different. The difference is that intents allow for a lot more flexibility. For our payment flow, however, using basically the same flow as before allows for minimal changes and keeps the process simple. The difference in the current implementation is that the Charge is contained in the payment intent.

The states pictured in the link is all handled by Stripe automatically, we shouldn't need to take action in any other case than if the intent failed or succeeded. i.e for us the flow is as simple as:
1. creating a payment intent on the server
2. attaching a payment method on the client
4. We monitor the state with webhooks,  and check whether the payment failed or succeeded.

If we use the `payment_failed` [webhook event](https://stripe.com/docs/payments/payment-intents/verifying-status#handling-webhook-events) to determine if the payment failed, we should have pretty much the same flow, and it should remove the need for migrating since we still only use the single `charge`, and as far as I see, there is no reason for the `PAYMENT_FAILURE` to match Stripe?
This, of course, does not fix the issue with the old payments, and may not be the best solution, but it does work just fine.

lego/apps/events/tasks.py
orhanhenrik: Maybe make the field nullable if stripe returns null/undefined?

lego/apps/events/tasks.py
orhanhenrik: Should we notify user or retry if event_type is not succeeded?

lego/apps/events/views.py
```diff
@@ -167,12 +168,8 @@ def administrate(self, request, *args, **kwargs):
         event_data = populate_event_registration_users_with_grade(event_data)
         return Response(event_data)
 
-    @decorators.action(
-        detail=True, methods=["POST"], serializer_class=StripeTokenSerializer
-    )
+    @decorators.action(detail=True, methods=["POST"], serializer_class=BaseSerializer)
```
orhanhenrik: Since the endpoint has breaking changes you need to deploy both frontend and backend at the same time and at a time when nobody uses the site. Otherwise a users payment might fail if you deploy while they are trying to pay.

You could also add a new endpoint and remove this one later to avoid breaking changes.

lego/apps/events/views.py
```diff
@@ -167,12 +168,8 @@ def administrate(self, request, *args, **kwargs):
...
     def payment(self, request, *args, **kwargs):
```
orhanhenrik: Should we rename this to something more descriptive? Since it's no longer really a payment

lego/apps/events/tasks.py
```diff
@@ -138,16 +138,19 @@ def async_unregister(self, registration_id, logger_context=None):
 
 
 @celery_app.task(base=Payment, bind=True)
-def async_payment(self, registration_id, token, logger_context=None):
+def async_payment(self, registration_id, logger_context=None):
```
orhanhenrik: Maybe rename this since it is no longer really doing a payment?

lego/apps/events/serializers/registrations.py
```diff
@@ -144,8 +144,8 @@ class Meta:
         read_only = True
 
 
-class StripeTokenSerializer(serializers.Serializer):
-    token = serializers.CharField()
+class StripePaymentIntentSerializer(serializers.Serializer):
```
orhanhenrik: Is this used anywhere?

lego/apps/events/views.py
odinuge: We have to deploy this _before_ the next payment, so that should be no problem. (as no payments will go through now)

lego/apps/events/constants.py
odinuge: > The flow actually isn't that much different.

It is still different. A lot of the phrasing doesn't make sense with the new API (ref. comments from @orhanhenrik). I guess we can draw a simple chart on Wednesday, so we all agree on the flow, and on how it should be implemented.

> there is no reason for the PAYMENT_FAILURE to match Stripe?

Neh, in theory no, as long as we make sure that we always convert the values.

> This, of course, does not fix the issue with the old payments, and may not be the best solution, but it does work just fine.

Ofc, but it is related to the code this PR changes :man_shrugging:

odinuge:
Also, I suspect that the tests fail with this change, so those have to be fixed too.

lego/apps/events/tasks.py
LudvigHz: This is handled in the frontend, as stripe serves the errors and we handle retrying there. The webhooks are really just for setting the correct values in the db.

lego/apps/events/serializers/registrations.py
LudvigHz: yes, its used in the webhook task

lego/apps/events/serializers/registrations.py
```diff
@@ -135,26 +135,42 @@ class Meta:
...
+
+class StripePaymentIntentSerializer(serializers.Serializer):
+    id = serializers.CharField()
+    amount = serializers.IntegerField()
+    status = serializers.CharField()
+    metadata = StripeMetaSerializer()
+    last_payment_error = StripePaymentIntentErrorSerializer(allow_null=True)
+
+
+class StripeChargeSerializer(serializers.Serializer):
```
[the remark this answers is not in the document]
LudvigHz: This is used for webhooks that don't give the entire `payment intent`, but only the `charge`. Specifically, this is used for the `charge.refunded` webhook event.

lego/apps/events/tasks.py
```diff
@@ -137,64 +147,169 @@ def async_unregister(self, registration_id, logger_context=None):
...
             metadata={
                 "EVENT_ID": event.id,
                 "USER_ID": self.registration.user.id,
                 "USER": self.registration.user.full_name,
                 "EMAIL": self.registration.user.email,
             },
         )
-        log.info("stripe_payment_success", registration_id=self.registration.id)
-        return response
-    except stripe.error.CardError as e:
```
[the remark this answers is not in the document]
LudvigHz: We don't handle any card errors explicitly anymore since all of the requests to Stripe that contain card data is handled on the client. The only payment error handling we do on the backend is monitoring the webhook event and setting the payment as failed, and the details are handled on the client.

lego/apps/events/tasks.py
```diff
@@ -210,11 +325,19 @@ def check_for_bump_on_pool_creation_or_expansion(self, event_id, logger_context=
...
 @celery_app.task(serializer="json", bind=True, base=AbakusTask)
 def stripe_webhook_event(self, event_id, event_type, logger_context=None):
+    """
+    Task that handles webhook events from Stripe, and updates the users registration in accordance
+    with the payment status.
+    """
     self.setup_logger(logger_context)
 
-    if event_type in ["charge.failed", "charge.refunded", "charge.succeeded"]:
+    if event_type in ["payment_intent.payment_failed", "payment_intent.succeeded"]:
```
odinuge: I would prefer if these were constants and not hardcoded here.

lego/apps/events/tasks.py
```diff
@@ -137,64 +147,169 @@ def async_unregister(self, registration_id, logger_context=None):
...
+
+    if registration.payment_intent_id is None:
+        log.error(
+            "Attempting to cancel a payment intent that does not exist",
+            registration=registration.id,
+        )
+        raise ValueError("Payment intent does not exist")
+
+    try:
+        stripe.PaymentIntent.cancel(registration.payment_intent_id)
```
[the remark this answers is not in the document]
LudvigHz: Hmm, yeah I didn't think about that :fearful:

lego/apps/events/views.py
```diff
@@ -167,31 +169,34 @@ def administrate(self, request, *args, **kwargs):
...
-        serializer = self.get_serializer(data=request.data)
-        serializer.is_valid(raise_exception=True)
         event_id = self.kwargs.get("pk", None)
         event = Event.objects.get(id=event_id)
         registration = event.get_registration(request.user)
 
         if not event.is_priced or not event.use_stripe:
-            raise PermissionDenied()
+            raise PermissionDenied()  # TODO better response
```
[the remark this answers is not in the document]
LudvigHz: :ok:

lego/apps/events/tasks.py
```diff
@@ -224,11 +347,32 @@ def stripe_webhook_event(self, event_id, event_type, logger_context=None):
         if not registration:
             log.error("stripe_webhook_error", event_id=event_id, metadata=metadata)
             raise WebhookDidNotFindRegistration(event_id, metadata)
-        registration.charge_id = serializer.data["id"]
-        registration.charge_amount = serializer.data["amount"]
-        registration.charge_amount_refunded = serializer.data["amount_refunded"]
-        registration.charge_status = serializer.data["status"]
-        registration.save()
+
+        if event_type == "charge.refunded":
```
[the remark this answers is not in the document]
LudvigHz: I can try to make it a bit cleaner :+1:

lego/apps/events/tasks.py
LudvigHz: :ok:

lego/apps/events/tasks.py
```diff
@@ -86,6 +84,16 @@ def async_register(self, registration_id, logger_context=None):
...
             )
         log.info("registration_success", registration_id=self.registration.id)
+        if (
+            self.registration.event.is_priced
+            and self.registration.event.use_stripe
+            and self.registration.pool is not None
+        ):
+            chain(
+                async_initiate_payment.s(registration_id),
+                save_and_notify_payment.s(registration_id),
```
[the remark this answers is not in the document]
LudvigHz: The same task is used when you click `Betal` and you receive the payment intent, so it's just to reuse code really.

lego/apps/events/views.py
LudvigHz: I couldn't really come up with anything better, and i feel like it makes sense, considering you have to post to the route in order to pay?

lego/apps/events/tasks.py
```diff
@@ -119,6 +127,8 @@ def async_unregister(self, registration_id, logger_context=None):
                     activation_time=activation_time,
                 )
             )
+        if registration.event.payment_intent_id:
+            async_cancel_payment.s(registration_id).delay()
```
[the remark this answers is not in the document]
hanswilw: You can write `.delay(registration_id)` directly here.

lego/apps/events/exceptions.py
```diff
@@ -11,6 +11,16 @@ class APIPaymentExists(APIException):
     default_detail = "Payment already exist."
 
 
+class APIPaymentDenied(APIException):
+    status_code = 403
+    default_detail = "The registration is not permittet to pay."
```
hanswilw: permitted

LudvigHz: :upside_down_face:

PR #1681, opened 2019-09-18
New Polls, Articles, Pages fixtures

SmithPeder:
- Creates a poll with 4 options that, and its pin on the frontpage
- Creates some new articles, and tag one of them as weekly

### DIFF
--- a/lego/apps/articles/fixtures/development_articles.yaml
+++ b/lego/apps/articles/fixtures/development_articles.yaml
@@ -8,6 +8,10 @@
 
 - model: tags.Tag
   fields:
+    tag: weekly
+
+- model: tags.Tag
+  fields:
     tag: sample-text
 
 - model: articles.Article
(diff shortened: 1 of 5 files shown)


lego/apps/articles/fixtures/development_articles.yaml
```diff
@@ -74,3 +78,65 @@
...
     youtube_url: https://www.youtube.com/watch?v=bLHL75H_VEM
+
+- model: articles.Article
+  pk: 4
+  fields:
+    title: Weekly - UKE 37
+    cover: test_article_cover.png
+    description: Takket være de flinke medlemmene i itDAGENE så fikk vi en pangstart
+      på uken med tre-retters middag.
+    text: Takket være de flinke medlemmene i itDAGENE så fikk vi en pangstart på
```
odinuge: Should this be html?

lego/apps/polls/fixtures/development_polls.yaml
```diff
@@ -13,20 +14,41 @@
...
+    name: Tinder
+    votes: 5
     poll: 1
 
 - model: polls.Option
   pk: 2
   fields:
-    name: Bombom
-    votes: 25
+    name: Grinder
```
kotlarz: offside 😨

kotlarz:
Could you add some fixtures for quote reactions as well? I forgot to request it in the PR.

lego/apps/articles/fixtures/development_articles.yaml
SmithPeder: The old fixtures are saved as normal text but I see that events save their `text` as html so we can do it here as well. Both get text from from the `Content` model, so it sould work just fine

lego/apps/polls/fixtures/development_polls.yaml
SmithPeder: So you're saying that **Grinder** is offside, but **Tinder** is okey?  I dont think Webkom should discriminate like this 🔥

kotlarz: Very true, I apologize to the Offside community.

JonasBak: Typo tho: https://www.grindr.com/

SmithPeder: 🕵

SmithPeder:
@kotlarz Where do `reactions` pop-up on quotes? On the detailed route?

orhanhenrik:
Should rebase so that github actions run

lego/apps/articles/fixtures/development_articles.yaml
```diff
@@ -15,16 +19,17 @@
...
+    text:
+      "<p>Lorem Ipsum er rett og slett dummytekst fra og for trykkeindustrien.\
+      \ Lorem Ipsum har vært bransjens standard for dummytekst helt siden 1500-tallet,\
+      \ da en ukjent boktrykker stokket en mengde bokstaver for å lage et prøveeksemplar\
+      \ av en bok. Lorem Ipsum har tålt tidens tann usedvanlig godt, og har i tillegg\
+      \ til å bestå gjennom fem århundrer også tålt spranget over til elektronisk\
+      \ typografi uten vesentlige endringer. Lorem Ipsum ble gjort allment kjent i\
+      \ 1960-årene ved lanseringen av Letraset-ark med avsnitt fra Lorem Ipsum, og\
+      \ senere med sideombrekkingsprogrammet Aldus PageMaker som tok i bruk nettopp\
+      \ Lorem Ipsum for dummytekst. </p>"
```
LudvigHz: Why all the backslashes? Seems a bit excessive to me :upside_down_face:

SmithPeder: Actually note sure, but that was how `<HTML>` was done in the other fixtures, so I just `cmd-c` + `cmd-v`

SmithPeder:
> Should rebase so that github actions run

I did, but I don't think it ran?

kotlarz:
> > Should rebase so that github actions run
> 
> I did, but I don't think it ran?

Yeah there's something odd there. There's no way to force a flow at the moment either. How about if you remove the "branches" restriction from the .github workflow file?

SmithPeder:
> Yeah there's something odd there. There's no way to force a flow at the moment either. How about if you remove the "branches" restriction from the .github workflow file?

I can try that, there is something odd about workflows on `force-push`.

SmithPeder:
Doesn't look like it triggered now either, and that was a normal push 😕

SmithPeder:
@orhanhenrik This now includes a fixture for the `for-bedrifter` page, so the Link can be tested now _(or whatever you were doing with the link)_

PR #1687, opened 2019-09-24
Upgrade camelcase package

orhanhenrik:
Also refactor tests to use `.json()` instead of `.data` so they get camelcased responses.

Tests can also post camelcased requests instead of underscored data to make it more similar to how the API is used. Both camelcase and underscore is supported though, because of how the camelcase package works.


lego/apps/users/tests/test_abakusgroup_api.py
```diff
@@ -172,7 +157,7 @@ def test_create_validate_permissions(self):
 
         response = self.client.post(_get_list_url(), group)
         self.assertEqual(response.status_code, 400)
-        self.assertEqual(expected_data, response.data)
+        self.assertEqual(expected_data, response.json())
```
odinuge: You should be able to remove the `@skip` here (hopefully :sweat_smile:)

PR #1771, opened 2020-01-28
Remove day limit on previous personal events in ICal

AdrianAndersen:
With this change, personal events will remain in the user's personal calendar (eg. iCal, Google Calendar) [Previously, personal events disappreard from the user's personal calendar after 10 days]

### DIFF
--- a/lego/apps/ical/viewsets.py
+++ b/lego/apps/ical/viewsets.py
@@ -93,7 +93,7 @@ class ICalViewset(viewsets.ViewSet):
         )
 
         permission_handler = get_permission_handler(Meeting)
-        meetings = permission_handler.filter_queryset(request.user)
+        meetings = permission_handler.filter_queryset(request.user, Meeting.objects.all())
 
         utils.add_events_to_ical_feed(feed, following_events)
         utils.add_meetings_to_ical_feed(feed, meetings)
--- a/lego/apps/ical/viewsets.py
+++ b/lego/apps/ical/viewsets.py
@@ -93,7 +93,7 @@ class ICalViewset(viewsets.ViewSet):
         )
 
         permission_handler = get_permission_handler(Meeting)
-        meetings = permission_handler.filter_queryset(request.user, Meeting.objects.all())
+        meetings = permission_handler.filter_queryset(request.user, Meeting.objects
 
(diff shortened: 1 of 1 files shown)


lego/apps/ical/viewsets.py
```diff
@@ -89,21 +89,11 @@ def personal(self, request):
...
 
         permission_handler = get_permission_handler(Meeting)
-        meetings = permission_handler.filter_queryset(
-            request.user,
-            Meeting.objects.filter(
-                end_time__gt=timezone.now()
-                - timedelta(days=constants.HISTORY_BACKWARDS_IN_DAYS)
-            ),
-        )
+        meetings = permission_handler.filter_queryset(request.user)
```
odinuge:
```suggestion
        meetings = permission_handler.filter_queryset(request.user, Meeting.objects.all())
```

lego/apps/ical/viewsets.py
```diff
@@ -89,21 +89,11 @@ def personal(self, request):
         permission_handler = get_permission_handler(Event)
         following_events = permission_handler.filter_queryset(
             request.user,
-            Event.objects.filter(
-                followers__follower_id=request.user.id,
-                end_time__gt=timezone.now()
-                - timedelta(days=constants.HISTORY_BACKWARDS_IN_DAYS),
-            ).all(),
+            Event.objects.filter(followers__follower_id=request.user.id).all(),
```
odinuge: If `constants.HISTORY_BACKWARDS_IN_DAYS` is no longer used (I am not sure tho), it should be removed from the constants file. Just search through the repo after `HISTORY_BACKWARDS_IN_DAYS`.

odinuge:
```suggestion
            Event.objects.filter(followers__follower_id=request.user.id),
```
All isn't required.

odinuge:
When you fix the issue I commented (see test output here: https://ci.abakus.no/webkom/lego/3603/7), the tests will run, and we will see if this breaks any of them

lego/apps/ical/viewsets.py
AdrianAndersen: This constant is still used for all events, thus it should remain.

Issue #1377, opened 2019-02-03
Dark theme


LudvigHz:
Add the option to enable global dark theme.

LudvigHz:
Fixed in https://github.com/webkom/lego-webapp/pull/1760

PR #1775, opened 2020-02-08
Dark Mode - Backend 

AdrianAndersen:
Add backend support for saving darkMode preference in user settings. Solves #1377


kotlarz:
Check out [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme). That way you won't have to implement anything in the back-end :+1:

AdrianAndersen:
Yeah, that would be nice to use as well. However, it would also be good to have dark mode as a user setting. This would enable synced theme settings across all user devices regardless of web browser settings. 
It might be possible to use a combo, where the default theme is decided by browser preference settings, while the user can alter this in their profile settings.

lego/apps/users/migrations/0022_user_selected_theme.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('users', '0021_auto_20190829_1632'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='user',
+            name='selected_theme',
+            field=models.CharField(blank=True, max_length=50, verbose_name='selected theme'),
```
kotlarz: Create a constant with the themes and set `choices=constants.THEMES`, name it something good.

lego/apps/users/migrations/0023_auto_20200212_0709.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.1.11 on 2020-02-12 07:09
```
kotlarz: Merge the migrations, you only need 1.

lego/apps/users/models.py
```diff
@@ -37,7 +37,8 @@
 
 class MembershipHistory(models.Model):
     user = models.ForeignKey("users.User", on_delete=models.CASCADE)
-    abakus_group = models.ForeignKey("users.AbakusGroup", on_delete=models.CASCADE)
+    abakus_group = models.ForeignKey(
```
kotlarz: Fix your linter (`black`).

lego/apps/users/serializers/users.py
```diff
@@ -18,7 +18,8 @@ class DetailedUserSerializer(serializers.ModelSerializer):
     abakus_groups = PublicAbakusGroupSerializer(many=True)
     past_memberships = PastMembershipSerializer(many=True)
     penalties = serializers.SerializerMethodField("get_valid_penalties")
-    profile_picture = ImageField(required=False, options={"height": 200, "width": 200})
+    profile_picture = ImageField(required=False, options={
```
kotlarz: Fix your linter (`black`).

PR #1778, opened 2020-02-11
Add fixture for the new survey type template

finnss:


lego/apps/surveys/fixtures/survey_template_fixtures.yaml
```diff
@@ -54,6 +54,13 @@
     event: 8
     template_type: 'event'
 
+- model: surveys.Survey
+  pk: 9
+  fields:
+    title: 'Template: alternate'
+    event: 9
+    template_type: 'alternate'
```
JonasBak: I think it might be `alternative_presentation`, but I might be wrong.

SmithPeder: Yes

mrfarstad:
LGTM 💯

PR #1779, opened 2020-02-11
Mail notification when there is less than one hour until event registration opens

theodorsm:
When a user follows an event, the following mail will be sent between 30 and 60 min before registration opens.

Screenshot_20200211_235743

Feedback needed.


theodorsm:
@erlingfn Thanks! I will look into it.

theodorsm:
> I think it might safer to do it the [survey way](https://github.com/webkom/lego/blob/4fee87bc34e0d08a30518d33db7f0d1b94de91b4/lego/apps/surveys/tasks.py#L20) by adding a `notified/sent` field to `FollowEvent` that you can filter on.

> You should probably also make sure that you only get one mail when you qualify for more than one pool for the same event

I agree that the survey way would be safer, and I will add a test to ensure you only recive one mail.

JonasBak:
If you do this (the survey way) you should also remember to check if the user is already registered or in the waiting list for the event, as you automatically follow an event you register for.

lego/apps/followers/tasks.py
```diff
@@ -0,0 +1,40 @@
...
+        activation_date__lte=timezone.now() + timedelta(minutes=60),
+    ).prefetch_related("event", "event__followers", "event__followers__follower")
+
+    for pool in pools:
+        for followsevent in pool.event.followers.all():
+            user = followsevent.follower
+            if (
+                pool.permission_groups.filter(
+                    id__in=[user.id for user in user.all_groups]
+                ).exists()
```
LudvigHz: This is a bit confusing, as you use `user` instead of `group` when iterating over the users groups. Rename this so it makes more sense :smile:

lego/apps/followers/models.py
```diff
@@ -25,6 +25,7 @@ class FollowEvent(Follower):
     target = models.ForeignKey(
         "events.Event", related_name="followers", on_delete=models.CASCADE
     )
+    sent = models.BooleanField(default=False)
```
LudvigHz: Maybe rename this to something more understandable as well. f.ex `registration_notification_sent` or something shorter :stuck_out_tongue_closed_eyes:

PR #1788, opened 2020-02-21
Add the field displayType to define graph selection

iselinee:
To be able to save what type of graph you choose, in order to represent results, displayType was added.


yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
JonasBak: Probably remove this (as well as the `node_modules/.yarn-integrity`)? :trolleybus:

PR #1794, opened 2020-02-24
Company Interest + BDB company connection

finnss:
Implements several updates to company-interest and BDB as per requests from bedkom:

- Adds functionality for selecting registered BDB companies when submitting a company interest form. If the company has not been registered previously, submitting with any name is still supported too. When a BDB company is chosen, the appropriate semester in the BDB table is updated to "Interested".
- Adds phone number as mandatory fields to company-interest.


JonasBak:
> Adds functionality for selecting registered BDB companies when submitting a company interest form.

If I understand correctly, this imply that we'll leak what companies are in the bdb? Isn't that a bit unfortunate? We should at least ask Bedkom explicitly if this is ok.

finnss:
Yeah

> > Adds functionality for selecting registered BDB companies when submitting a company interest form.
> 
> If I understand correctly, this imply that we'll leak what companies are in the bdb? Isn't that a bit unfortunate? We should at least ask Bedkom explicitly if this is ok.

Yeah, true. I asked now, let's see what they say.

SmithPeder:
@JonasBak We discussed it on the Slack channel, where I pointed out the same issue. We had a discussion on an itDAGENE meeting with the same issue. In my opinion its a very minor issue. We have almost all (relevant) companies in the database already so you can't use it so see "What companies Abakus target" or something. I agree that is Bedkom agrees to it (as they requested it) we can just rol with it.

lego/apps/companies/models.py
```diff
@@ -106,9 +106,25 @@ class Meta:
 
 
 class CompanyInterest(PersistentModel, TimeStampModel):
-    company_name = models.CharField(max_length=255)
+    """
+    Either company_name OR company should be null. We prefer to use a Foreign key to
+    an existing company when making a Company Interest, because that lets us
+    do things like update the BDB status for that company to "interested".
+    It still needs to be possible to register interested without being in the BDB,
```
SmithPeder: interest*

lego/apps/companies/models.py
```diff
@@ -106,9 +106,25 @@ class Meta:
...
+    however, in which case company is Null and the string company_name is used instead.
```
SmithPeder: in this*

kotlarz:
> Look legal to me

Looks*

finnss:
> The code looks good, yet we should hold back until we have resolved the issue with other people being able to see which companies are interested.

We received a reply, posted in Slack. We should be good to go.

PR #1798, opened 2020-02-29
Add initial support for displaying permissions for users and groups

odinuge:


lego/apps/users/models.py
```diff
@@ -222,6 +228,20 @@ def past_memberships(self):
             user=self, abakus_group__deleted=False
         ).select_related("abakus_group")
 
+    @abakus_cached_property
+    def nested_permissions(self):
```
odinuge: Could probably do this cleaner, but this is a good first try :tm:

odinuge:
Agree @SmithPeder!

Did a small rewrite now, so you can take a look.

SmithPeder:
Nice, those names made a bit more sense 🥇

odinuge:
Ill give this a try in master :tm:

PR #1799, opened 2020-02-29
Add support for fetching descendant memberships

odinuge:
~impossible to do this using django-filter, so this was the simplest way
to do it.

When fetching with descendants=true, it will return all memberships for
the group and for all the child groups.


odinuge:
lgtm

Issue #1801, opened 2020-03-02
Can't delete profile picture, only update it by uploading another one


techiag:
abakus.no/users/me

PR #1807, opened 2020-03-11
Add funcionality to remove profile picture

theodorsm:
Needed for issue: https://github.com/webkom/lego/issues/1801

Frontend: https://github.com/webkom/lego-webapp/pull/1795


theodorsm:
> it would be much simpler to simply set the profile_picture to None. This way we can also change the default picture without updating any users.
> If you change the field `picture = FileField(related_name="user_pictures")` to also have `blank=True`, I think you will be able to update your own user from the frontend by setting `profilePicture` to `null`.

I agree this is a much cleaner way of doing it, will try it out.

lego/apps/users/models.py
```diff
@@ -371,6 +371,10 @@ def email_address(self):
     def profile_picture(self, value):
         self.picture = value
 
+    def profile_picture_remove(self):
+        self.profile_picture = None
+        self.save()
+
```
orhanhenrik: I think you can remove this as well

lego/apps/users/tests/test_models.py
```diff
@@ -109,6 +109,26 @@ def test_set_profile_picture(self):
         self.user.save()
         self.assertEqual(self.user.profile_picture, "abakus.png")
 
+    def test_remove__male_profile_picture(self):
```
orhanhenrik: I don't think you need these tests either, since you're not really adding any functionality, but only changing a field to nullable

lego/apps/users/models.py
```diff
@@ -13,7 +13,7 @@
 
 from lego.apps.events.constants import PRESENT
 from lego.apps.external_sync.models import GSuiteAddress, PasswordHashUser
-from lego.apps.files.models import FileField
+from lego.apps.files.models import File, FileField
```
orhanhenrik: Don’t need this extra import

lego/apps/users/tests/test_users_api.py
```diff
@@ -359,6 +359,16 @@ def test_update_not_failing_when_not_student(self):
         user = self.all_users.exclude(student_username__isnull=False).first()
         self.successful_update(user, user)
 
+    def test_update_to_remove_profile_picture(self):
+        """Try to remove profile picture by updating user with profilePIcture equals null"""
+        user = self.all_users.exclude(student_username__isnull=False).first()
```
orhanhenrik: Should also check that user already has a picture, so that the change actually has an effect

PR #1859, opened 2020-04-27
Healthchecks

SmithPeder:
# Add healthcheck endpoint

There was an unregistered `health` app, but I removed this and created a new one. Migrated over the permission schema from health, as it's smart to limit the IP range that can use the endpoint.

The `/healthcheck` will now only return `200 OK` if all the checks work.

## This change also does the following
- Sort `lego/settings/base.py` INSTALLED APPS
- Set `auto-reload` logger to info and not debug

image


lego/settings/logging.py
```diff
@@ -64,6 +64,7 @@ def skip_if_testing(*args, **kwargs):
             "level": "INFO",
             "propagate": False,
         },
+        "django.utils.autoreload": {"level": "INFO"},
```
LudvigHz, line 67: :heart_eyes:

LudvigHz:
Is there a reason we don't use this in the Dockerfile? https://docs.docker.com/engine/reference/builder/#healthcheck

SmithPeder:
This `healthcheck` can only run in the `uwsgi` based containers. So, if you add a `native` healthcheck to the image it will fail for all celery workers. These need their own `celery inspect ping` command to check their health.

Issue #1920, opened 2020-08-31
Error with backend regex for fileupload


magnhan:
So there are still some errors with the backend-regex.
It's either:
- Numbers
- Spaces
- Punctuation

SmithPeder:
What are the filenames?

magnhan:
"17. mai kvadrat.jpg" and "17. mai kvadrat.png"

SmithPeder:
Can you rename them to something like `asdfasdf.jpg` and try again 👍 
There are some problems with the `regex`, either on numbers, spaces, punctuation or something else 🙃

magnhan:
That worked!

PR #1923, opened 2020-09-07
Make the regex more flex

SmithPeder:
Fixes: https://github.com/webkom/lego/issues/1920

The regex will not match (.) periods correctly, and people have a habit of using that in their filenames.

Now the regex will match on \w multiple times with (.) before the final filetype.
image


odinuge:
`make fixme`

SmithPeder:
i cry

LudvigHz (approved):
lgtm

PR #1935, opened 2020-09-22
[Ignore for now]Add cookieConsent to User model

theodorsm:
Related to: https://github.com/webkom/lego-webapp/pull/1948


odinuge (changes requested):
As discussed this has to be connected to a version, so that we can ask again when we update the stuff

SmithPeder (changes requested):
As Odin says,  there should be a **version** that you consent to.

I don't know what the best way to make this "easy" but maybe something like this:

Instead of a boolean for the consent store a string with the **latest** version that they consent, with the default being an empty string.

Then this could be the easy schema as an example:
```js
" " == does not consent
"v1_23_09_2020" == consent to v1 created at 23.sep
"v2_23_12_2020" == consent to v2 created at 23.des
```

Then the backend can return the `cookie_version` with the site_meta like this
image

It's kinda manual since we have to edit the backend when we create a new version
image

But then, in the frontend, instead of just checking 
> cookie == true

you can check

> is the users consented version the same as the site meta cookie version

And when you POST the consent, you can post the version from the site meta.

I think it's quite elegant and simple 🤷 but I'd love to hear if someone has an easier solution 👍

odinuge:
Just as a ref., this is what NRK says: `Forhåndsinnstilling i nettleser om at brukeren aksepterer informasjonskapsler anses som samtykke. Det er tilstrekkelig at brukeren samtykker én gang for det samme formålet. Brukeren skal til enhver tid ha mulighet til å trekke tilbake sitt samtykke.`


https://www.nrk.no/retningslinjer/informasjonskapsler-_cookies_-1.11109868

theodorsm:
Closed by https://github.com/webkom/lego-webapp/pull/1985

Issue #1958, opened 2020-09-28
CompanyInterest submission validation error


SmithPeder:
So the interest form is broken in prod, so our updated changes won't work (as the normal form does not work).

![image](https://user-images.githubusercontent.com/23152018/94412783-82fa1f00-017a-11eb-8d97-32905b3fc2b0.png)

![image](https://user-images.githubusercontent.com/23152018/94412622-48908200-017a-11eb-9ed7-559a1ac7ebaa.png)

```
Traceback (most recent call last):
  File "/Users/smith/code/abakus/lego/venv/lib/python3.7/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/Users/smith/code/abakus/lego/venv/lib/python3.7/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/Users/smith/code/abakus/lego/venv/lib/python3.7/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/smith/code/abakus/lego/venv/lib/python3.7/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/Users/smith/code/abakus/lego/venv/lib/python3.7/site-packages/rest_framework/viewsets.py", line 116, in view
    return self.dispatch(request, *args, **kwargs)
  File "/Users/smith/code/abakus/lego/venv/lib/python3.7/site-packages/rest_framework/views.py", line 495, in dispatch
    response = self.handle_exception(exc)
  File "/Users/smith/code/abakus/lego/venv/lib/python3.7/site-packages/rest_framework/views.py", line 455, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/Users/smith/code/abakus/lego/venv/lib/python3.7/site-packages/rest_framework/views.py", line 466, in raise_uncaught_exception
    raise exc
  File "/Users/smith/code/abakus/lego/venv/lib/python3.7/site-packages/rest_framework/views.py", line 492, in dispatch
    response = handler(request, *args, **kwargs)
  File "/Users/smith/code/abakus/lego/venv/lib/python3.7/site-packages/rest_framework/mixins.py", line 21, in create
    self.perform_create(serializer)
[...]

PR #1960, opened 2020-09-28
Disable notifications for CompanyInterest

SmithPeder:
Solves: https://github.com/webkom/lego/issues/1958


erlingfn (approved):
This looks good. Removing notifications is not an ideal solution, but the most important thing is that the company interest form actually works

PR #1972, opened 2020-10-09
Add drone conf to deploy lego

SmithPeder:
Do the same as we are doing for `lego-webapp` by using `drone` to deploy to staging and production.
image


LudvigHz (changes requested):
LGTM! Just add a signature as in lego-webapp as we have a lot of secrets available here now.

LudvigHz (approved):

PR #1979, opened 2020-10-13
Add Backend Support for Improved Photo Consents

AdrianAndersen:
Adds backend support for adding and modifying photo consents per semester.
image

Depends on [#1997](https://github.com/webkom/lego-webapp/pull/1997)


lego/apps/users/views/users.py
```diff
@@ -173,3 +174,25 @@ def change_grade(self, request, *args, **kwargs):
...
+            return Response(status=status.HTTP_403_FORBIDDEN)
+
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        user = self.get_object()
+
+        with transaction.atomic():
+            photo_consent = user.photo_consents.get(
+                semester=request.data.get("semester"), domain=request.data.get("domain")
+            )
```
LudvigHz, line 195: What happens when the photo_consent does not exist?

lego/apps/users/serializers/users.py
```diff
@@ -173,6 +208,29 @@ def get_valid_penalties(self, user):
         serializer = PenaltySerializer(instance=qs, many=True)
         return serializer.data
 
+    def get_photo_consents(self, user):
+        pc = PhotoConsent.objects.filter(user=user)
+
+        from datetime import datetime
```
LudvigHz, line 214: :scream:
Also: `django.utils.timezone`

lego/apps/users/serializers/users.py
```diff
@@ -173,6 +208,29 @@ def get_valid_penalties(self, user):
...
+                semester=current_semester,
+                domain="SOCIAL_MEDIA",
+                is_consenting=None,
+            )
+            PhotoConsent.objects.create(
+                user=user,
+                semester=current_semester,
+                domain="WEBSITE",
+                is_consenting=None,
+            )
```
LudvigHz, line 230: I think it makes more sense to put this logic somewhere else, f.ex on the model or user? Especially since this code is duplicated. And the .count() is a little ambiguous. Maybe check if each specific photo_consent exists?

LudvigHz (commented):
Remember to squash your migrations and remove the dev stuff (from the commits themselves!) when you are finished :smile:.

lego/apps/events/models.py
```diff
@@ -783,11 +782,6 @@ class Registration(BasisModel):
     presence = models.CharField(
         max_length=20, default=constants.UNKNOWN, choices=constants.PRESENCE_CHOICES
     )
-    photo_consent = models.CharField(
```
odinuge, line 786: Are we sure we want to nuke old consents? Doesn't that conflict with the whole purpose?

AdrianAndersen: Yeah, true dat. Keeping them as legacy for now, so that it is easy to remove in like 6 yrs :S. :+1:

lego/apps/users/views/users.py
AdrianAndersen: There has to be photo consents in order for the user to send a request to change them. (front end) However, I added validation and creation of a photo consent here in the wierd case this is not generated beforehand. :shrug:

lego/apps/users/serializers/users.py
AdrianAndersen: Yup, fixed this :sweat_smile:

lego/apps/users/serializers/users.py
AdrianAndersen: Moved this logic to the PhotoConsent Model. Changed count to instead filter out the individual photo consents, and create new ones if needed. :+1:

lego/apps/events/models.py
```diff
@@ -266,6 +267,13 @@ def register(self, registration):
         if self.use_contact_tracing and user.phone_number is None:
             raise NoPhoneNumber()
 
+        current_semester = "autumn" if self.start_time.month > 7 else "spring"
```
LudvigHz, line 270: Use constants here.

lego/apps/users/models.py
```diff
@@ -522,3 +531,48 @@ def ignore_date(date):
         elif winter_to < (date.month, date.day) < winter_from:
             return False
         return True
+
+
+class PhotoConsent(BasisModel):
+    user = models.ForeignKey(
+        User, related_name="photo_consents", on_delete=models.CASCADE
+    )
+    semester = models.CharField(max_length=6)
```
LudvigHz, line 540: This should be constrained with some choices. As we have upgraded to django 3 now, you can also use the new enumeration types provided by Django[1]

[1]: https://docs.djangoproject.com/en/3.2/ref/models/fields/#enumeration-types

lego/apps/users/models.py
```diff
@@ -522,3 +531,48 @@ def ignore_date(date):
...
+    year = models.PositiveIntegerField()
+    domain = models.CharField(choices=constants.PHOTO_CONSENT_DOMAINS, max_length=100)
+    is_consenting = models.BooleanField(blank=True, null=True, default=None)
+
+    def get_consents(self, user):
+        photo_consents = PhotoConsent.objects.filter(user=user)
+        now = timezone.now()
+        current_semester = "autumn" if now.month > 7 else "spring"
```
LudvigHz, line 548: Use constants here as well

lego/apps/users/models.py
```diff
@@ -522,3 +531,48 @@ def ignore_date(date):
...
+        current_year = now.year
+        if not any(
+            pc.year == current_year
+            and pc.semester == current_semester
+            and pc.domain == "SOCIAL_MEDIA"
```
LudvigHz, line 553: And here

lego/apps/users/models.py
```diff
@@ -522,3 +531,48 @@ def ignore_date(date):
...
+            for pc in photo_consents
```
LudvigHz, line 554: This probably works, but is very slow and unnecessary. Do all this as the DB level by creating a proper filter query and use `<Queryset>.exists()` instead of the for loop.

lego/apps/users/models.py
```diff
@@ -522,3 +531,48 @@ def ignore_date(date):
...
+            and pc.domain == "WEBSITE"
+            for pc in photo_consents
+        ):
+            PhotoConsent.objects.create(
+                user=user,
+                year=current_year,
+                semester=current_semester,
+                domain="WEBSITE",
+                is_consenting=None,
+            )
```
LudvigHz, line 576: Same here of course

lego/apps/users/migrations/0031_auto_20220111_0943.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 2.2.24 on 2022-01-11 09:43
```
LudvigHz, line 1: Could you squash this migration with the previous one?

lego/apps/users/tests/test_models.py
```diff
@@ -381,3 +381,43 @@ def test_penalty_offset_is_calculated_correctly(self, mock_now):
...
+        self.assertEqual(pc.user, self.test_user)
+        self.assertEqual(pc.year, self.current_year)
+        self.assertEqual(pc.semester, self.current_semester)
+        self.assertEqual(pc.domain, "SOCIAL_MEDIA")
+        self.assertIsNone(pc.is_consenting)
+
+        pc.is_consenting = True
+        pc.save()
+
+        self.assertEquals(pc.is_consenting, True)
```
LudvigHz, line 423: I would strongly prefer some more tests here. f.ex. Test get consent permissions. Test that the a "new" consent is created when getting consents, that the default values are correct. Test setting different values etc.

lego/apps/users/views/users.py
```diff
@@ -173,3 +175,33 @@ def change_grade(self, request, *args, **kwargs):
...
+
+        with transaction.atomic():
+            photo_consent = user.photo_consents.get(
+                year=serializer.validated_data["year"],
+                semester=serializer.validated_data["semester"],
+                domain=serializer.validated_data["domain"],
+            )
+
+            if not request.user.has_perm(EDIT, user):
+                raise ValueError("Cannot update other user's consent")
```
LudvigHz, line 198: Any reason for fetching the photo consent before doing the permission check?

LudvigHz (changes requested):
Nice work. Just some small things here and there and we're good! :shipit:

lego/apps/users/views/users.py
AdrianAndersen: Not really. Fixed

LudvigHz:
@AdrianAndersen Is this ready for review again?

lego/apps/users/tests/test_models.py
AdrianAndersen: Done

AdrianAndersen:
> @AdrianAndersen Is this ready for review again?

It has been a long while, but I finally got around to create some more tests for this. 🐌

So this PR and the frontend one is now complete. 🎉 

I would really appreciate it if you find the time to re-review this. 🕵🏻

lego/apps/users/tests/test_users_api.py
```diff
@@ -476,3 +477,97 @@ def test_own_penalties_serializer(self, mock_now):
...
+                {
+                    "user": self.test_user.id,
+                    "year": self.current_year,
+                    "semester": self.current_semester,
+                    "domain": WEBSITE_DOMAIN,
+                    "isConsenting": False,
+                },
+            )
+            self.assertEqual(response.status_code, 403)
+        except ValueError:
```
theodorsm, line 567: Why is this try/except needed?

lego/apps/users/models.py
```diff
@@ -526,3 +535,43 @@ def ignore_date(date):
...
+
+
+class PhotoConsent(BasisModel):
+    user = models.ForeignKey(
+        User, related_name="photo_consents", on_delete=models.CASCADE
+    )
+    semester = models.CharField(max_length=6, choices=constants.SEMESTERS)
+    year = models.PositiveIntegerField()
+    domain = models.CharField(choices=constants.PHOTO_CONSENT_DOMAINS, max_length=100)
+    is_consenting = models.BooleanField(blank=True, null=True, default=None)
```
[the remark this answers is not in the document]
AdrianAndersen: I think it should be null by default. This is because we want the users to take a stance on photo consent. I think it is VERY likely that if we default this to False, we will never be able to take pictures in any event in the future. (Since few will actually go out of their way to consent)

I am not sure about the None-bugs you are referring to, but in this case, I think it is handled ok. The alternative is to have a separate field of whether they have taken stance or not, but I think this might complicate it too much. 

Let me know what you think tho.

lego/apps/users/tests/test_users_api.py
AdrianAndersen: This is because the api throws an error when you are not authorised to change the consent. I am not quite sure try/except is the best and cleanest way to handle this in the tests, but let me know if you know a better way

lego/apps/users/models.py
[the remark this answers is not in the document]
theodorsm: Okay, that makes sense, I agree.

lego/apps/users/views/users.py
```diff
@@ -176,3 +176,35 @@ def change_grade(self, request, *args, **kwargs):
...
+        serializer_class=PhotoConsentSerializer,
+    )
+    def update_photo_consent(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        user = self.get_object()
+
+        with transaction.atomic():
+            if not request.user.has_perm(EDIT, user):
+                raise ValueError("Cannot update other user's consent")
```
theodorsm, line 193: This should be not a regular python Error, but from the rest_framework.exceptions, like this: 

`
raise PermissionDenied(detail="Cannot update other user's consent")
`

Using PermissionDenied we do not have to try/except it elsewhere in our codebase + not stop our server from running. In addition, this will respond with the correct HTTP status code (403) and with the detail as a message.

lego/apps/users/tests/test_users_api.py
theodorsm: The try/except can be removed when the raised error is changed to PermissionDenied.

lego/apps/users/tests/test_users_api.py
```diff
@@ -476,3 +477,97 @@ def test_own_penalties_serializer(self, mock_now):
...
+                self.test_user_url,
+                {
+                    "user": self.test_user.id,
+                    "year": self.current_year,
+                    "semester": self.current_semester,
+                    "domain": WEBSITE_DOMAIN,
+                    "isConsenting": False,
+                },
+            )
+            self.assertEqual(response.status_code, 403)
```
theodorsm, line 566: `
self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
`

lego/apps/users/tests/test_users_api.py
```diff
@@ -476,3 +477,97 @@ def test_own_penalties_serializer(self, mock_now):
...
+            self.test_user_url,
+            {
+                "user": self.test_user.id,
+                "year": self.current_year,
+                "semester": self.current_semester,
+                "domain": SOCIAL_MEDIA_DOMAIN,
+                "isConsenting": False,
+            },
+        )
+        self.assertEqual(response.status_code, 200)
```
theodorsm, line 543: `
        self.assertEqual(response.status_code, status.HTTP_200_OK)
`

lego/apps/users/tests/test_users_api.py
```diff
@@ -476,3 +477,97 @@ def test_own_penalties_serializer(self, mock_now):
...
+                "domain": WEBSITE_DOMAIN,
+                "isConsenting": True,
+            },
+        )
+        updated_consent = self.test_user.photo_consents.get(
+            year=self.current_year,
+            semester=self.current_semester,
+            domain=WEBSITE_DOMAIN,
+        )
+        self.assertEqual(response.status_code, 200)
```
theodorsm, line 520: `
        self.assertEqual(response.status_code, status.HTTP_200_OK)
`

theodorsm (changes requested):

lego/apps/users/tests/test_users_api.py
AdrianAndersen: done

lego/apps/users/tests/test_users_api.py
AdrianAndersen: done

lego/apps/users/views/users.py
AdrianAndersen: Thank you for the extensive suggestion! 😄 This makes a lot of sense. I have implemented them now. 👍🏻

AdrianAndersen: Thank you for the extensive suggestion! 😄 This makes a lot of sense. I have implemented them now. 👍🏻

lego/apps/users/tests/test_users_api.py
```diff
@@ -476,3 +477,95 @@ def test_own_penalties_serializer(self, mock_now):
...
+                "isConsenting": False,
+            },
+        )
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+        consent = self.test_user.photo_consents.get(
+            year=self.current_year,
+            semester=self.current_semester,
+            domain=WEBSITE_DOMAIN,
+        )
+        self.assertEqual(consent.is_consenting, None)
```
theodorsm, line 571: Could use assertIsNone(), but that is nitpicking.

theodorsm (approved):
Excited to get this into production! Great work :100:

lego/apps/users/tests/test_users_api.py
AdrianAndersen: Done ✅

lego/apps/users/views/users.py
```diff
@@ -176,3 +176,35 @@ def change_grade(self, request, *args, **kwargs):
...
+
+            photo_consent = None
+            try:
+                photo_consent = user.photo_consents.get(
+                    year=serializer.validated_data["year"],
+                    semester=serializer.validated_data["semester"],
+                    domain=serializer.validated_data["domain"],
+                )
+            except Exception:
+                pass
```
LudvigHz, line 203: eeeh what?
What is this meant to catch? And why are you just `pass`ing it?
If you need it, please catch the actual exception class and not the general one.

lego/apps/users/views/users.py
```diff
@@ -176,3 +176,35 @@ def change_grade(self, request, *args, **kwargs):
...
+
+            if photo_consent is None:
+                PhotoConsent.objects.create(**serializer.validated_data)
+            else:
+                photo_consent.is_consenting = serializer.validated_data["is_consenting"]
+                photo_consent.save()
```
LudvigHz, line 209: This seems like you want [`update_or_create()`](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#update-or-create)

LudvigHz (changes requested):
Awesome!
Just fix up some small things and clear up some questions and we're good to go!

lego/apps/users/models.py
```diff
@@ -446,6 +446,15 @@ def number_of_penalties(self):
...
         return count or 0
 
+    def has_registered_photo_consents_for_semester(self, event_year, event_semester):
+        consents = PhotoConsent.objects.filter(
+            user=self, year=event_year, semester=event_semester
+        )
+        for consent in consents:
+            if not isinstance(consent.is_consenting, bool):
+                return False
+        return True
```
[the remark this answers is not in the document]
AdrianAndersen: Yeah, it is always created, so this should not be a problem. Rewrote this into a pure query also.

lego/apps/users/constants.py
```diff
@@ -90,6 +90,22 @@
...
+    (AUTUMN, AUTUMN),
+)
+
+WEBSITE_DOMAIN = "WEBSITE"
+SOCIAL_MEDIA_DOMAIN = "SOCIAL_MEDIA"
+PHOTO_CONSENT_DOMAINS = (
+    (WEBSITE_DOMAIN, WEBSITE_DOMAIN),
+    (SOCIAL_MEDIA_DOMAIN, SOCIAL_MEDIA_DOMAIN),
+)
+
```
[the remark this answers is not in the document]
AdrianAndersen: Hey, that's cool, but I decided not to rewrite this time around. 😉

lego/apps/users/models.py
```diff
@@ -526,3 +535,43 @@ def ignore_date(date):
...
+            )
+
+        if not website_consent.exists():
+            PhotoConsent.objects.create(
+                user=user,
+                year=current_year,
+                semester=current_semester,
+                domain=constants.WEBSITE_DOMAIN,
+                is_consenting=None,
+            )
```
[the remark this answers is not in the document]
AdrianAndersen: True! Really cool 😄

lego/apps/users/views/users.py
AdrianAndersen: Yeaaaah. That was a stupid fix for a test run. Removed altogether because I use update_or_create().

lego/apps/users/views/users.py
AdrianAndersen: Yup! Thanks!

lego/apps/users/views/users.py
```diff
@@ -176,3 +176,28 @@ def change_grade(self, request, *args, **kwargs):
...
+        serializer_class=PhotoConsentSerializer,
+    )
+    def update_photo_consent(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        user = self.get_object()
+
+        with transaction.atomic():
+            if not request.user.has_perm(EDIT, user):
+                raise PermissionDenied(detail="Cannot update other user's consent")
```
LudvigHz, line 193: You can remove this transaction now.

LudvigHz (approved):
This looks good now!
Nice job, let's get this out ASAP!

lego/apps/events/serializers/registrations.py
```diff
@@ -49,8 +49,8 @@ class RegistrationCreateAndUpdateSerializer(BasisModelSerializer):
         required=False, choices=constants.PAYMENT_STATUS_CHOICES
     )
     presence = PresenceField(required=False, choices=constants.PRESENCE_CHOICES)
-    photo_consent = ConsentField(
-        required=False, choices=constants.PHOTO_CONSENT_CHOICES
+    LEGACY_photo_consent = ConsentField(
+        required=False, choices=constants.LEGACY_PHOTO_CONSENT_CHOICES
```
LudvigHz, line 53: You can remove this, we don't need (or want) users to create or update legacy consents :grimacing:

lego/apps/events/serializers/registrations.py
```diff
@@ -59,7 +59,7 @@ class Meta:
             "id",
             "feedback",
             "presence",
-            "photo_consent",
+            "LEGACY_photo_consent",
```
LudvigHz, line 62: Remove this as well

lego/apps/users/views/users.py
AdrianAndersen: Removed

lego/apps/events/serializers/registrations.py
AdrianAndersen: true. Removed

lego/apps/events/serializers/registrations.py
AdrianAndersen: Yup. Done ✅

PR #1980, opened 2020-10-13
Third party consent

erlingfn:
Depends on [#2001](https://github.com/webkom/lego-webapp/pull/2001)

Add backend support for third party consent
Add optional phone number to user model


lego/apps/events/serializers/events.py
```diff
@@ -284,6 +296,20 @@ def validate(self, data):
...
                     }
                 )
+        if self.instance is not None and "share_info_flag" in data:
+            pools = self.instance.pools.all()
+            reg_time = min(pool.activation_date for pool in pools)
+            if data[
+                "share_info_flag"
+            ] != self.instance.share_info_flag and reg_time < datetime.now(
+                timezone.utc
+            ):
```
LudvigHz, line 306: You can do this a lot easier with a single query:
```python
self.instance.pools.filter(activation_time__lte=timezone.now()).exists()
```

Also, you can use `timezone.now()` instead of `datetime(...)` to get the correct timezone.

LudvigHz (changes requested):
`...SerializerExport` -> `...ExportSerializer`. We ahould keep the convention of having the `Serializer` postfix on all serializer names.

Cool, just some small things :100:

lego/apps/users/serializers/users.py
```diff
@@ -85,7 +85,16 @@ class AdministrateUserSerializer(PublicUserSerializer):
...
+        fields = PublicUserSerializer.Meta.fields + ("abakus_groups", "allergies",)
+
+
+class AdministrateUserSerializerExport(PublicUserSerializer):
+    """
+    Used by the events app when listing user registrations, and has permission to export user data.
+    """
+
+    class Meta(PublicUserSerializer.Meta):
+        fields = AdministrateUserSerializer.Meta.fields + ("email",)
```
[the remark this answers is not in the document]
erlingfn: Your assumption is correct. We discussed this exact problem on site earlier this week, so it is your call @LudvigHz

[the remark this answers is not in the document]
LudvigHz: I agree with @odinuge, (and what you said while we discussed this @erlingfn :upside_down_face:). I am pretty sure it is always the person responsible for the event that creates it, so having the permission scheme be that just the creator is allowed is a nice solution imo.

Having a limit on the amount of days after the event is not a bad idea either. I think something like 30 days should be sufficient?

lego/apps/events/models.py
```diff
@@ -253,6 +255,9 @@ def register(self, registration):
         if self.registration_close_time < current_time:
             raise EventHasClosed()
 
+        if self.share_info_flag and user.phone_number is None:
```
odinuge, line 258: Should make some logic to handle this and get the error msg to the client, and show it. This will currently fail bad, and the user will keep seeing the loading indicator. (same applies to the other errors as well tho.)

Maybe outside the scope of this tho.

lego/apps/events/models.py
```diff
@@ -86,6 +87,7 @@ def registration_close_time(self):
     youtube_url = CharField(
         max_length=200, default="", validators=[youtube_validator], blank=True
     )
+    share_info_flag = models.BooleanField(default=False)
```
odinuge, line 90: Have we settled on `share_info_flag`? This is essentially _only_ for smittesporing, and not a general `share-info`-flag

lego/apps/events/serializers/events.py
```diff
@@ -223,9 +227,17 @@ class Meta(EventReadSerializer.Meta):
             "unregistered",
             "waiting_registrations",
             "use_consent",
+            "share_info_flag",
+            "created_by",
         )
 
 
+class EventAdministrateExportSerializer(EventAdministrateSerializer):
```
odinuge, line 235: Is this a `noop` by design, or am I missing something?

lego/apps/events/serializers/events.py
```diff
@@ -223,9 +227,17 @@ class Meta(EventReadSerializer.Meta):
             "unregistered",
             "waiting_registrations",
             "use_consent",
+            "share_info_flag",
+            "created_by",
```
odinuge, line 231: This should just return a `pk`?

lego/apps/events/serializers/pools.py
```diff
@@ -55,6 +56,10 @@ class PoolAdministrateSerializer(PoolReadAuthSerializer):
     registrations = RegistrationReadDetailedSerializer(many=True)
 
 
+class PoolAdministrateExportSerializer(PoolReadAuthSerializer):
```
odinuge, line 59: This is supposed to be a `noop` as well?

And isn't it better to extend `PoolAdministrateSerializer` instead?

lego/apps/events/serializers/registrations.py
```diff
@@ -152,6 +153,10 @@ class Meta:
         read_only = True
 
 
+class RegistrationReadDetailedExportSerializer(RegistrationReadDetailedSerializer):
```
odinuge, line 156: Another `noop`?

odinuge: Ahh, nvm. different user serializer

lego/apps/events/serializers/pools.py
odinuge: Ahh, nvm. different user serializer

lego/apps/events/serializers/events.py
odinuge: Ahh, nvm. different user serializer

lego/apps/events/models.py
erlingfn: What about `share_for_conact_tracing` or something along those lines?

LudvigHz: maybe `use_contact_tracing`?

lego/apps/events/models.py
AdrianAndersen: Should be fixed to all cases. Thinking this is not that important to fix here since the admitt button is hidden in the frontend anyways.

lego/apps/events/serializers/events.py
AdrianAndersen: yes

lego/apps/events/serializers/events.py
```diff
@@ -227,8 +227,8 @@ class Meta(EventReadSerializer.Meta):
             "unregistered",
             "waiting_registrations",
             "use_consent",
-            "share_info_flag",
-            "created_by",
+            "use_contact_tracing",
+            "created_by",UG
```
LudvigHz, line 231: :thinking:

lego/apps/users/models.py
```diff
@@ -10,6 +10,7 @@
 
 from mptt.fields import TreeForeignKey
 from mptt.models import MPTTModel
+from phonenumber_field.modelfields import PhoneNumber, PhoneNumberField
```
LudvigHz, line 13: Unused import?

lego/apps/events/serializers/events.py
LudvigHz: You still think we only should have the user id, and not the entire public user?

LudvigHz (commented):
Looks good!
Just fix the last small stuff and lets get this merged!

lego/apps/events/serializers/events.py
AdrianAndersen: Ooo, my typo. Yeah, I only use this to compare the user id with the user who created the event.

lego/apps/users/models.py
AdrianAndersen: true dat

LudvigHz (approved):
Nicely done!

PR #1989, opened 2020-10-19
Update filtering to work properly

odinuge:


lego/apps/users/fields.py
```diff
@@ -27,6 +27,19 @@ def __init__(self, field_kwargs=None, **kwargs):
...
 
 
+class PublicUserWithGroupsField(serializers.PrimaryKeyRelatedField):
+    def use_pk_only_optimization(self):
+        return False
+
+    def to_representation(self, value):
+        from lego.apps.users.serializers.users import (
+            PublicUserWithAbakusGroupsSerializer,
+        )
```
LudvigHz, line 37: Any reason not to import this at the top of the file?

LudvigHz (approved):
LGTM

lego/apps/users/fields.py
odinuge: Just copy pasta from the other one, but I suspect it is because of circular deps of some kind, but not 100% sure tho

LudvigHz: Jup, checks out :ok_hand:

PR #2032, opened 2021-01-22
Change name of RevyStyret group

erlingfn:


lego/settings/lego.py
```diff
@@ -39,7 +39,7 @@
     "Ordenen",
     "PR-ansvarlige",
     "Revy",
```
SmithPeder, line 41: you can delete this also

SmithPeder (changes requested):

SmithPeder (approved):
(y)

PR #2035, opened 2021-01-26
Add support for meeting desc

odinuge:


lego/apps/meetings/models.py
```diff
@@ -22,6 +22,7 @@ class Meeting(BasisModel):
     location = models.CharField(max_length=255)
     start_time = models.DateTimeField()
     end_time = models.DateTimeField(blank=True, null=True)
+    description = models.TextField(blank=True, default="")
```
odinuge, line 25: Or should we create a new "zoom" url field? I think this might age better after corona...

And, this will not highlight links at abakus.no with current impl (someone can fix), but it will in ical at google calendar.

LudvigHz: I think it's better to handle the links frontend, and keep it as a textfield. It makes it a bit simpler and should work just fine. Another option is to use the location, but that might be a bit strange?

odinuge: > I think it's better to handle the links frontend, and keep it as a textfield

Agree. Since ical is not fancy html-stuff.

>  Another option is to use the location, but that might be a bit strange?

Kinda. But gcal doesn't linkify links in the location fields of ical. This also allows for other info in the ical field.

odinuge: Can revert https://github.com/webkom/lego-webapp/commit/98c4219fe220c6b8b4d450e211e27ac13cebc45b and add some `target="blank"` stuff etc. if people want clickable links.

LudvigHz: yeah, that looks like a good solution :ok_hand:

LudvigHz (approved):

LudvigHz:
Seeing as the frontend version of this (https://github.com/webkom/lego-webapp/pull/2100) happened to already be deployed, we should maybe get this out ASAP as currently, I suspect meetings will be broken :upside_down_face:.

ingraso (approved):
:rocket:

PR #2047, opened 2021-02-07
Add support for negative values for reg limits

odinuge:


lego/apps/events/websockets.py
```diff
@@ -19,8 +19,11 @@ def find_event_groups(user):
     """
     Find all channels groups the user belongs to as a result
     of being signed up to future events.
+
+    Since we use websockets for registrations and event description uptdates, include events
```
SmithPeder, line 23:
```suggestion
    Since we use websockets for registrations and event description updates, include events
```

SmithPeder (changes requested):

lego/apps/events/websockets.py
SmithPeder: 😄

SmithPeder (approved):

LudvigHz (approved):

PR #2064, opened 2021-02-23
Add last_login to users

theodorsm:
The field `last_login` is inherited from Django AbstractBaseUser model. We need this field to be updated for **ALL** users if we want to _delete users as discussed_. 

Since we did not call `update_last_login` on JWT authentication for users earlier, some users might not have a `last_login` date (the field is set to NULL). **This PR contains a basic script for updating the `last_login` for ALL users to the current time.** As we don't use the field anywhere this should be safe. (?) We can use this script to ensure all users at least have a _last_login_ date from the time running the script and onwards.

The plan is to make another script for notifying all users that have not been logged in for the last X days that they are soon to have their account deleted. When the users log in again they will not be deleted when we run a separate delete inactive users for the X last days script.


lego/utils/management/commands/bump_last_login.py
```diff
@@ -0,0 +1,46 @@
...
+        new_time = timezone.now()
+        log.info(
+            '{:3d} users ready to update last login to "{}"'.format(
+                all_users.count(), new_time
+            )
+        )
+
+        if not is_dryrun:
+            for user in all_users:
+                self.update_last_login(user, new_time)
```
LudvigHz, line 46: This works, but I think it would be a lot better to just do this with one query. aka something like:
```python
User.objects.all().update(...)
```

This will change from ~16000 queries to just one.

LudvigHz (changes requested):
Nice. Script could be improved a bit though :+1:

lego/utils/management/commands/bump_last_login.py
theodorsm: Yes, this is MUCH better! Fixed in lastest commit

LudvigHz (approved):

lego/utils/management/commands/bump_last_login.py
```diff
@@ -0,0 +1,45 @@
...
+
+from django.utils import timezone
+
+from lego.apps.users.models import User
+from lego.utils.management_command import BaseCommand
+
+log = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
```
odinuge, line 11: Is this maybe better suited as a data migration instead? Then everyone gets it, including when running locally.

Is there a use case of this after we run it, or?

odinuge (commented):
Nice!

Have you checked if auth via oauth update this as well?

odinuge:
I think some people might use ldap and/or gmail, but maybe not use normal auth. Can  we get data from there as well? if the `n` is big enough, eg. a year, that might not be necessary tho.

lego/utils/management/commands/bump_last_login.py
theodorsm: I have been thinking and I am really struggling to find a use case for running this multiple times. A data migration is the way to go.

odinuge (approved):
+isort lgtm

LudvigHz:
@theodorsm Is this still WIP?

theodorsm:
> @theodorsm Is this still WIP?
No, I think this is ready to merge.

LudvigHz (approved):
LGTM

PR #2072, opened 2021-03-02
Fix postgres search

LudvigHz:
Improve postgres search _alot_ by using PostgreSQL's full text search. Also implement the `search` method to allow more comprehensive searching. This is now used when pressing enter in the search bar in the webapp. IMO it's a _massive_ improvement since it searches in the content again.

Currently this adds a query for all search indices (some still missing). I have though about making this more general, since a lot of the autocomplete queries are very similar, especially the `SearchQuery` for autocompletion. It's not a big deal though, but let me know if anyone has any thoughts! :hand:.

**TODO:**
- [x] Edit the rest of the search indices
- [x] Consider generalizing the `search` and `autocomplete` methods with a general method for the base `SearchIndex` class. (see below)
- [ ] ~~Consider adding pagination (this will probably be another PR, not really in scope for this)~~

I ended up adding a method for `search` and `autocomplete` that uses a `search_fields` and `autocomplete_fields` to create a search vector and query. This makes the code a lot smaller, and allows us to add a new search index by just adding the fields that should be used for search/autocomplete as attributes.

### DIFF
--- a/lego/apps/users/search_indexes.py
+++ b/lego/apps/users/search_indexes.py
@@ -22,6 +22,8 @@ class UserIndex(SearchIndex):
     )
     autocomplete_result_fields = ("username", "full_name", "profile_picture")
 
+    autocomplete_fields = ("first_name", "last_name", "username")
+
     def get_autocomplete(self, instance):
         return [
             instance.username,
@@ -29,16 +31,6 @@ class UserIndex(SearchIndex):
             instance.last_name,
             instance.first_name,
         ]
-
-    def autocomplete(self, query):
-        return self.queryset.annotate(
-            search=SearchVector("first_name", "last_name", "username")
-        ).filter(
-            search=SearchQuery(
-                ":* & ".join(query.split() + [""]).strip("& ").strip(),
-                search_type="raw",
-            )
-        )
 
 
 register(UserIndex)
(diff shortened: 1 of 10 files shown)


LudvigHz:
This is deployed in staging atm if anyone wants to test it out.

odinuge (commented):
nice, cool, awesome, 10/10

:shipit:

lego/apps/users/search_indexes.py
```diff
@@ -31,12 +32,12 @@ def get_autocomplete(self, instance):
...
+            search=SearchVector("first_name", "last_name", "username")
         ).filter(
-            Q(first_name__istartswith=query)
-            | Q(last_name__istartswith=query)
-            | Q(username__istartswith=query)
-            | Q(fullname__istartswith=query)
+            search=SearchQuery(
+                ":* & ".join(query.split() + [""]).strip("& ").strip(),
+                search_type="raw",
+            )
```
LudvigHz, line 40: This should have some better sorting :thinking: I'm thinking just `created_at` maybe is good? or maybe membership count or something (if that's not too discriminating?)

PeterJFB (approved):
Awesome stuff! Lgtm^^

PR #2075, opened 2021-03-04
Add email lists to MeSerializers

odinuge:


lego/apps/email/serializers.py
```diff
@@ -12,6 +12,23 @@
...
+    class Meta:
+        model = EmailList
+        fields = (
+            "id",
+            "users",
+            "name",
+            "email",
+            "groups",
+            "group_roles",
+            "require_internal_address",
```
LudvigHz, line 28: Do you need all these fields? I have a feeling the `/me` response is getting quite large now :upside_down_face:.

LudvigHz (approved):
Nice :100:

lego/apps/email/serializers.py
odinuge: Nah, not that important i guess. But they are quite small however, so waaay smaller than memberships and groups...

odinuge: Also, this drops `/users/me` from 20 to 15 queries in dev, so it is probably faster now.

odinuge: Can probably move stuff out of `/me` to separate endpoints, but perms are pretty nice when using it, and we still "need" all the info when looking at the user profile on the webapp.

odinuge: Also, this only return ids, and not complete user and group entities, so fairly small.

odinuge (commented):
Lets test in staging

lego/apps/email/serializers.py
LudvigHz: Hmm, yeah we prob have larger endpoints, so I guess it's fine :+1:

PR #2081, opened 2021-03-08
Fix bad equality check in survey permission handler

LudvigHz:
Ints <= 256 are preallocated on the heap, so any int under 257 points to
the same object in memory. Since the is operator compares object
references, two vars with the same value > 256 are allocated as
different memory objects meaning is will evaluate to False.

Changing to == fixes this issue by comparing the value instead.

see https://medium.com/peloton-engineering/the-dangers-of-using-is-in-python-f42941124027


odinuge (approved):
:shipit: Good catch!

PR #2085, opened 2021-03-09
Add unsubscribe link in mails

### LINKED ISSUE #2084: Emails should contain a link to edit email preferences
Should be to just add a link to the base email template, as well as add a link to mails sent by restricted.

amalieeh:
Fix #2084

Add a link to notification settings in bottom of mail where user can unsubscribe mails.
image

I also removed the facebook-link as it can increase the chances of the mail being reported as spam.
I changed the `<strong>`-tag with a `<b>`-tag because older clients support it.
I also deleted some comments that didn't seem necessary, but I dont really know... 

And for restricted mail.

### DIFF
--- a/lego/templates/email/base.html
+++ b/lego/templates/email/base.html
@@ -216,11 +216,6 @@
         <table cellspacing="0" cellpadding="0" width="100%" bgcolor="#ffffff" background="http://s3.amazonaws.com/swu-filepicker/4E687TRe69Ld95IDWyEg_bg_top_02.jpg" style="background-color:transparent">
           <tr>
             <td width="100%" height="80" valign="top" style="text-align: center; vertical-align:middle;">
-            <!--[if gte mso 9]>
-            <v:rect xmlns:v="urn:schemas-microsoft-com:vml" fill="true" stroke="false" style="mso-width-percent:1000;height:80px; v-text-anchor:middle;">
-              <v:fill type="tile" src="http://s3.amazonaws.com/swu-filepicker/4E687TRe69Ld95IDWyEg_bg_top_02.jpg" color="#ffffff" />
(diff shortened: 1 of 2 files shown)


lego/templates/email/base.html
```diff
@@ -216,27 +216,17 @@
         <table cellspacing="0" cellpadding="0" width="100%" bgcolor="#ffffff" background="http://s3.amazonaws.com/swu-filepicker/4E687TRe69Ld95IDWyEg_bg_top_02.jpg" style="background-color:transparent">
           <tr>
             <td width="100%" height="80" valign="top" style="text-align: center; vertical-align:middle;">
-            <!--[if gte mso 9]>
-            <v:rect xmlns:v="urn:schemas-microsoft-com:vml" fill="true" stroke="false" style="mso-width-percent:1000;height:80px; v-text-anchor:middle;">
-              <v:fill type="tile" src="http://s3.amazonaws.com/swu-filepicker/4E687TRe69Ld95IDWyEg_bg_top_02.jpg" color="#ffffff" />
-              <v:textbox inset="0,0,0,0">
-            <![endif]-->
```
LudvigHz, line 223: This code applies to anyone using outlook. I'm not sure what it does myself, but I'm guessing there is a reason we have it :sweat_smile:. Maybe someone with outlook can test?

LudvigHz (commented):
Awesome!

lego/templates/email/base.html
amalieeh: Nice spot! I though it was only some old comments left in the code.😅

I see no point deleting it now since it might have a function.

LudvigHz: Could you readd this, so we can merge this? :smile:

LudvigHz (approved):

PR #2092, opened 2021-03-16
Legacy registation count

theodorsm:
Add legacy count to events to keep count after users are deleted.


iselinee:
Looks good.

lego/apps/users/models.py
```diff
@@ -377,6 +377,17 @@ def get_default_picture(self):
         else:
             return "default_other_avatar.png"
 
+    def delete(self, using=None, force=False):
+        from lego.apps.events import constants
+        from lego.apps.events.models import Event
+
+        if force:
+            for event in Event.objects.all():
+                if event.is_admitted(self):
```
LudvigHz, line 386: :open_mouth: :snail: 
```python
Event.objects.filter(registrations__user=self)
```
or something. The above filter at least gets only events where the user is registered. You could prob add a filter for admission as well :thinking:.

```python
Event.objects.filter(Q(registrations__user=self) & Q(registrations__pool__isnull=False))
```
Might work. (some quick testing seems good).

lego/apps/users/models.py
```diff
@@ -377,6 +377,17 @@ def get_default_picture(self):
...
+                    if event.registrations.get(user=self).presence == constants.PRESENT:
```
LudvigHz, line 387: Again, you could do `event.registrations.filter(user=self, presence=constants.PRESENT).exists()` to avoid db lookups. (not really important here though)

lego/apps/events/models.py
```diff
@@ -721,6 +726,11 @@ def announcement_lookup(self):
         registrations = self.registrations.filter(status=constants.SUCCESS_REGISTER)
         return [registration.user for registration in registrations]
 
+    def add_legacy_registration(self):
+        constants.PRESENT,
```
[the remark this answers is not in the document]
theodorsm: Forgot to remove, fixed in latest commit

lego/apps/users/models.py
theodorsm: See _Oh yes, speed._ commit. 🏎️

lego/apps/users/models.py
theodorsm: Is this still relevant to do in latest commmit?

LudvigHz: nope. Looks good!

LudvigHz (approved):

lego/apps/events/models.py
```diff
@@ -652,7 +653,10 @@ def total_capacity(self):
     @property
     def registration_count(self):
         """ Prefetch friendly counting of registrations for an event. """
-        return sum([pool.registrations.all().count() for pool in self.pools.all()])
+        return sum(
```
odinuge, line 656: nit: The second parameter is the "start value" of the sum, so it might  be easier to just add the numbers together instead.
```suggestion
        return self.legacy_registration_count + sum(
```

odinuge (approved):
lgtm

Good work! :smile:

PR #2093, opened 2021-03-16
Add functionality to give view permission to event for specific groups

iselinee:


lego/apps/events/models.py
```diff
@@ -692,22 +692,18 @@ def waiting_registrations(self):
...
-        if abakom_only:
+    def set_groups_with_view_permission(self, groups_with_view_permission):
+        if len(groups_with_view_permission) > 0:
+            self.groups_with_view_permission.clear()
             self.require_auth = True
-            self.can_view_groups.add(abakom_group)
+            for group in groups_with_view_permission:
+                abakus_group = AbakusGroup.objects.get(name=group)
+                self.groups_with_view_permission.add(abakus_group)
+                self.can_view_groups.add(abakus_group)
```
LudvigHz, line 702: These two fields seem to be very similar. What is the difference? Do you need the `groups_with_view_permission` field?

lego/apps/events/serializers/events.py
```diff
@@ -102,6 +103,8 @@ class EventReadDetailedSerializer(TagSerializerMixin, BasisModelSerializer):
     text = ContentSerializerField()
     created_by = PublicUserSerializer()
 
+    groups_with_view_permission = AbakusGroupNameSerializer(many=True)
```
LudvigHz, line 106: Why not use `AbakusGroupListField`?

LudvigHz (changes requested):
Cool. I don't know exactly what issues we had with event object permissions before, but I definitely think this need a lot of tests. That is, with different groups, and tests for view permission, registration permission, view registrations, keyword permission overrides etc.

lego/apps/events/migrations/0031_event_groups_with_view_permission.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.13 on 2021-03-16 17:54
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("users", "0025_user_phone_number"),
```
theodorsm, line 9: See https://github.com/webkom/lego/commit/19cd30e01883a33c2d9153d63b8e9a9454a2277f, This should be updated to the latest migration.

theodorsm (changes requested):

lego/apps/events/models.py
```diff
@@ -50,7 +50,7 @@ class Event(Content, BasisModel, ObjectPermissionsModel):
     )
     location = models.CharField(max_length=100)
     cover = FileField(related_name="event_covers")
-    groups_with_view_permission = models.ManyToManyField(AbakusGroup, blank=True)
+    can_view_groups = models.ManyToManyField(AbakusGroup, blank=True)
```
LudvigHz, line 53: This is already included in the ObjectPermissionModel this model inherits from, so you don't need to add this :smile:.

lego/apps/events/models.py
```diff
@@ -692,17 +692,15 @@ def waiting_registrations(self):
     def waiting_registration_count(self):
         return self.waiting_registrations.count()
 
-    def set_groups_with_view_permission(self, groups_with_view_permission):
-        if len(groups_with_view_permission) > 0:
-            self.groups_with_view_permission.clear()
+    def set_can_view_groups(self, can_view_groups):
```
LudvigHz, line 695: I think this method can be removed now, since this only really sets `require_auth` :thinking:. (Might need it to set it still if there is no easier method though)

lego/apps/events/models.py
iselinee: Seems I was a bit too quick with just replacing groups_with_view_permission. Thanks for spotting it!

PR #2129, opened 2021-03-26
Make it possible to have an empty list of roles

SmithPeder:


odinuge (approved):

odinuge:
Should probably fix these for all ArrayFields tho., in order to avoid any issues...

PR #2130, opened 2021-03-28
Fix link, again

ivarn01:
I'm not sure why the link changes, but at least this fixes it. Perhaps we should find a better way of getting the image?


LudvigHz:
@ivarn01 This is most likely due to the recent upgrade of webpack, that probably changed how the hash is generated. The proper way to fix it will be to modify the webpack config here: https://github.com/webkom/lego-webapp/blob/cf19e70ba0c12dcde47b0ab06ec02de1fcd82de7/config/webpack.client.js#L219-L220

See the docs for how to fix this: https://webpack.js.org/guides/asset-modules/#custom-output-filename

The proper way to handle images is to hash, but since we don't have that many images, it should be fine to use the filename, maybe with a has of the filename or similar.

PR #2161, opened 2021-04-15
Fix blank group permissions field

LudvigHz:
Due to a recent change in Django-REST, fields need to be set `blank=True` for the serializer not to
invalidate when fields are omitted.

Also changes the name of Fondsstyret group to LDAP sync. (:pray: hope this doesn't break something).


SmithPeder (approved):
Have you renamed the group as well? Or du you plan do it after this change?

SmithPeder:
__missing_migrations__ 🙃

LudvigHz:
> Have you renamed the group as well? Or do you plan do it after this change?

I planned to to it after. It's not possible without this anyways due to the blank field issue.

PR #2171, opened 2021-04-27
Add permissions based on event type for event create update

LudvigHz:
This allows the use of keyword permissions for create/update actions and
limiting to event type.

F.ex. /sudo/admin/events/create/social allows creating _only_ events
with the `social` type. (You need the same with `edit` to edit social
events).

### TODO
- [ ] Tests
- [ ] Modify serializers to limit fields you can set based on the event_type (might be another PR)
- [ ] Modify administrator view to limit what data the event admin can see. (^^^^)

### DIFF
--- a/lego/apps/events/permissions.py
+++ b/lego/apps/events/permissions.py
@@ -1,7 +1,12 @@
+from rest_framework.permissions import BasePermission
+
 from structlog import get_logger
 
+from lego.apps.permissions.actions import action_to_permission
 from lego.apps.permissions.constants import CREATE, DELETE, EDIT, VIEW
+from lego.apps.permissions.models import ObjectPermissionsModel
 from lego.apps.permissions.permissions import PermissionHandler
+from lego.apps.permissions.utils import get_permission_handler
 
 log = get_logger()
 
@@ -9,6 +14,23 @@ log = get_logger()
 class EventPermissionHandler(PermissionHandler):
 
     perms_without_object = [CREATE, "administrate"]
+
+    def event_type_keyword_permissions(self, event_type, perm):
+        from lego.apps.events.models import Event
+
+        return self.keyword_permission(Event, perm) + "{event_type}/".format(
+            event_type=event_type
+        )
+
+    def has_event_type_level_permission(self, user, request, perm):
+        if request is None:
+            return True
+        event_type = request.data.get("event_type")
+
+        required_keyword_permissions = self.event_type_keyword_permissions(
+            event_type, perm
+        )
+        return user.has_perm(required_keyword_permissions)
 
 
 class RegistrationPermissionHandler(PermissionHandler):
@@ -31,7 +53,7 @@ class RegistrationPermissionHandler(PermissionHandler):
         obj=None,
         queryset=None,
         check_keyword_permissions=True,
-        **kwargs
+        **kwargs,
     ):
 
         is_self = self.is_self(perm, user, obj)
@@ -41,3 +63,14 @@ class RegistrationPermissionHandler(PermissionHandler):
         return super().has_perm(
             user, perm, obj, queryset, check_keyword_permissions, **kwargs
         )
+
+
+class EventTypePermission(BasePermission):
+    def has_permission(self, request, view):
+        from lego.apps.events.models import Event
+
+        perm = action_to_permission(view.action)
+        if perm in [CREATE, EDIT]:
+            handler = get_permission_handler(Event)
+            return handler.has_event_type_level_permission(request.user, request, perm)
+        return super().has_permission(request, view)
--- a/lego/apps/events/views.py
+++ b/lego/apps/events/views.py
@@ -26,6 +26,7 @@ from lego.apps.events.exceptions import (
 )
 from lego.apps.events.filters import EventsFilterSet
 from lego.apps.events.models import Event, Pool, Registration
+from lego.apps.events.permissions import EventTypePermission
 from lego.apps.events.serializers.events import (
     EventAdministrateExportSerializer,
     EventAdministrateSerializer,
@@ -76,6 +77,8 @@ class EventViewSet(AllowedPermissionsMixin, viewsets.ModelViewSet):
     )
     ordering_fields = ("start_time", "end_time", "title")
     ordering = "start_time"
+
+    permission_classes = (EventTypePermission,)
 
     def get_queryset(self):
         user = self.request.user


lego/apps/users/fixtures/initial_abakus_groups.yaml
```diff
@@ -3,143 +3,462 @@
 #
 - model: users.abakusgroup
   pk: 1
-  fields: {deleted: false, name: Users, description: "Brukere p\xE5 Abakus.no", contact_email: '',
```
odinuge, line 6: Move this to separate pr pls

lego/apps/events/permissions.py
```diff
@@ -10,6 +16,38 @@ class EventPermissionHandler(PermissionHandler):
...
+
+    def has_event_type_level_permission(self, user, request, perm):
+        """
+        Check if a user has the required event_type
+        permissions for a certain action.
+
+        Return `True` if the user has permission to perform the request
+        """
+        if request is None:
+            return True
```
odinuge, line 39: Is this ever used for something? My initial thought is that it should be `False`?

lego/apps/events/permissions.py
```diff
@@ -10,6 +16,38 @@ class EventPermissionHandler(PermissionHandler):
...
+        event_type = request.data.get("event_type")
+
+        # The request might be a patch without the event_type. This should be allowed
+        if event_type is None:
+            return True
```
odinuge, line 44: So if user tries to create event without `event_type`, it is allowed (from a permission perspective)?

lego/apps/users/fixtures/initial_abakus_groups.yaml
LudvigHz: This is just a change in how the yaml is generated. So really it only adds one group, which is needed for the tests.

lego/apps/events/permissions.py
odinuge: Also, return `True` here, does probably not necessary mean `[the] the user has permission to perform the request`

docs/events.rst
```diff
@@ -62,6 +62,33 @@ can join that pool is popped. This registration has it's ``waiting_list``, ``wai
...
+Events mostly uses the default permissions in LEGO, but has some custom permission handlers.
+
+Permissions based on event type
+*******************************
+For events we want to limit the different types of events a user can create or edit. So for
+events, one can add append the `event type` on the ``create`` and ``edit`` keyword permissions.
+F.ex: ``/sudo/admin/events/create/social/`` gives the user only access to create events with
+`event_type = social`.
+
+If a user has the ``/sudo/admin/events/edit/social/``, they have access to change an event object's type `to` ``social``.
```
odinuge, line 77: I don't think I see a reason for this one. If you can create one event with type x and y, you should also be able to change events from x to y, and from y to x.

To me a `/sudo/admin/events/edit/social/` permission looks like a permission to edit all events with type social.

Is there ever a reason where `can edit type` and `can create type` is not the same?

odinuge: If there is a good reason to split them, then it is ok, but I just can't see any reasons.. :thinking:

lego/apps/users/fixtures/initial_abakus_groups.yaml
odinuge: But why is the formatting changed?

odinuge: It's impossible to read the diff as is

docs/events.rst
LudvigHz: I see your point and I agree with the fact that changing from x to y and reverse should be possible. As of now, the user will need object permission to edit the event, and in addition they will need permission to create events with type y. 

I very much agree that this is weird, and I too would think /edit/social/ means editing social events. But in reality it means the user har permission to do an edit action with social type events.

I think checking create permissions make sense from a use perspective, but the reason it is like this is that it makes more sense since we find the required keyword permission based on the action (in this case `edit`). 

We could modify this to check create permissions for edit actions as well (which would remove the /edit/event_type/ permissions completely). I'm not really sure what makes more sense here though so if the consensus is the latter, I'm fine with changing it.

lego/apps/events/permissions.py
LudvigHz: Hmm, yes you're correct. Even though it shouldn't be possible it's more understandable if we enforce it for edit.

docs/events.rst
odinuge: Yeah, I see your point, and I definitely agree about the fact that it makes the code shorter and more generic.

My only concern is that there is _never_ a reason to give a user only one of the perms, and therefore it imo. is easier to just merge them.

The current impl works, so its fine. I just think it can save some possible confusion later.

odinuge: We probably only want to show types they are allowed to use as well, and then having a single perm per type makes it easier to do that in a "generic" way.

lego/apps/events/permissions.py
odinuge: Yeah. I read this as `oh, you don't change the evnet type, ofc. you have permission to do this action. Accepted`. It is however only applicable in case user has obj perm, and do not have `/events/edit/` perm. An no people should have that perm anyways, except for people with `sudo`.

docs/events.rst
odinuge: But big +1 for docs :raised_hands:

lego/apps/events/permissions.py
LudvigHz: I think this might be left over from when before I moved the permission check to the view. I'll check it out and remove/change it to `False`. Thanks :+1:

lego/apps/events/permissions.py
LudvigHz: I see. The thing is that this method is only meant to check the `event_type`. It does not care about object permissions. This method must be used in conjunction with the "default" permission handlers. As you see in the permission class used in the view. The request needs to pass **both** this and the default permissions (which enforce obj permissions).

I see the confusion though. I will specify this a bit more in the docs, and maybe in some comments so it is more understandable :+1:. And I will only do this check for `EDIT`.

odinuge (approved):

lego/apps/users/fixtures/initial_abakus_groups.yaml
LudvigHz: I'm guessing some dependency got updated. I moved the regeneration of the formatting to a separate commit now.

lego/apps/events/permissions.py
```diff
@@ -33,18 +33,27 @@ def has_event_type_level_permission(self, user, request, perm):
         Check if a user has the required event_type
         permissions for a certain action.
 
-        Return `True` if the user has permission to perform the request
+        **WARNING**: This method ONLY enforces keyword permissions for event types. It does not
```
odinuge, line 36: :+1:

PeterJFB (approved):
LGTM!

PR #2196, opened 2021-05-23
Add boolean field telling frontend to hide poll results

LarsSelbekk:
As requested by Glorious Leader @AdrianAndersen, this PR will enable a poll to hide the number of votes cast for each option. Note that because this is done in a hurry all the data will still be sent to the client, meaning the frontend will have to respect the new field for this to have any effect, and it can always be bypassed simply using browser Developer Tools. A PR for the frontend is incoming. Also, for the author to see the result they would have to either check the response object manually in for example DevTools, or make the result public for everyone by editing the poll.

Future work on this feature should include hiding the vote tally in the response object as well, perhaps from everyone but the administrators, or everyone but the poll author (which does currently not seem to be stored).


LudvigHz:
Nice one! :raised_hands:. However, I really think this should be done properly, and not just use the flag in the frontend. It think the best way is to remove `options` from the list serializer (I think it isn't used anywhere, but a quick check might be a good idea), and then we can do something like this in the viewset to hide the results:

```diff
class PollViewSet(AllowedPermissionsMixin, viewsets.ModelViewSet):
    queryset = Poll.objects.all()
    ordering = "-created_at"

    def get_serializer_class(self):
        if self.action == "create":
            return PollCreateSerializer
        if self.action in ["update", "partial_update"]:
            return PollUpdateSerializer
+        if self.action == "retrieve":
+			poll = self.get_object()
+			if poll.results_hidden:
+				return HiddenResultsDetailedPollSerializer
            return DetailedPollSerializer
        return PollSerializer
```

Where `HiddenResultsDetailedSerializer` just has `options` as well omitted.

LarsSelbekk:
Thanks @LudvigHz, it should work with the new commit. Took a while, but this seems to work, though I'm not sure whether it's the cleanest way (overriding to_representation in the serializers). I think the way you proposed is a little vulnerable to people adding new views and unintentionally exposing the data, though I don't have a lot of experience with Django.

Using this strategy the individual option tallies are completely hidden until the results are made public, meaning not even the original poll author can view the results before that. This could be enabled by saving which user authored a poll and only showing tallies to them, or perhaps showing to everyone with editing access, but from what @AdrianAndersen tells me that isn't needed for now.

LudvigHz:
I guess that works, but I wouldn't recommend it :sweat_smile: as it's a bit verbose and not really that understandable. I guess you would still need to see the options, but not votes. The proper method to do this would be to create a serializer for `Options` omitting the `votes` field, and using this on the serializer for `Polls` that should hide results.

>  I think the way you proposed is a little vulnerable to people adding new views and unintentionally exposing the data, though I don't have a lot of experience with Django.

Sure, but very very unlikely and not really an issue. I don't think anyone will be able to miss what data they expose when creating a view (not that we are going to add a new view for Polls anytime in the near future)

> Using this strategy the individual option tallies are completely hidden until the results are made public, meaning not even the original poll author can view the results before that.

I would recommend using the user and checking permissions as well when selecting the serializer (like I proposed above). (Although this is not critical, as you said)
So something like this should work:
```diff
    def get_serializer_class(self):
        if self.action == "create":
            return PollCreateSerializer
        if self.action in ["update", "partial_update"]:
            return PollUpdateSerializer
+        if self.action == "retrieve":
+			poll = self.get_object()
+			if poll.results_hidden and not request.user.has_perm(EDIT, poll):
+				return HiddenResultsDetailedPollSerializer
            return DetailedPollSerializer
        return PollSerializer
```

LarsSelbekk:
That was a very simple solution, but it's now possible to see every option tally by querying the complete list at /api/v1/polls. Do you have any idea whether it's difficult to show the votes for those the user are allowed to, but not others? To me it seems like the easiest solution would be to hide all tallies when getting the entire list, though that seems like a violation of REST principles.

LudvigHz:
> That was a very simple solution, but it's now possible to see every option tally by querying the complete list at /api/v1/polls. Do you have any idea whether it's difficult to show the votes for those the user are allowed to, but not others? To me it seems like the easiest solution would be to hide all tallies when getting the entire list, though that seems like a violation of REST principles.

Not including the vote count for list seems like the way to go imo. As explained, I don't think we use the vote count (or options for that matter) when fetching the list anywhere on the webapp :+1:. So just creating another serializer (preferably that is used as a base for the detail ones) should be fine. There are plenty of examples in other parts of the project. E.g. `events` for how this is done.

LarsSelbekk:
I didn't extract a base serializer because those serializers interact with so many things I have no idea of, I didn't dare. I created two extremely simple new serializers instead, and changed PollSerializer to hide option counts (affecting the list endpoint). Removing the options entirely caused more problems than I believe it is worth. I've tested quite a bit now, and everything seems fine. To see the result one has to first vote on the poll even as the author (unless perusing DevTools), however I believe amending that is out of scope for this feature.

LudvigHz:
> I didn't extract a base serializer because those serializers interact with so many things I have no idea of, I didn't dare. I created two extremely simple new serializers instead, and changed PollSerializer to hide option counts (affecting the list endpoint). 

That was exactly how I was thinking! :star_struck:. Nice job!

LudvigHz:
So another thing that probably was not too obvious is that we have a separate endpoint for all frontpage items (due to performance reasons), so you will need to add the Hiddden serializer here as well :sweat_smile: :
https://github.com/webkom/lego/blob/65be8c1f402b3d3c35badb010dd84878983f035f/lego/apps/frontpage/views.py#L97-L101

No need to care about creator permission check here, as long as they can access the vote count from the detail page, we're good :+1:.

<details><summary>Here is a working patch if you want:</summary>
<p>

```diff
diff --git a/lego/apps/frontpage/views.py b/lego/apps/frontpage/views.py
index 7b63d180..bf0fd581 100644
--- a/lego/apps/frontpage/views.py
+++ b/lego/apps/frontpage/views.py
@@ -11,7 +11,10 @@ from lego.apps.events.serializers.events import FrontpageEventSerializer
 from lego.apps.permissions.constants import LIST
 from lego.apps.permissions.utils import get_permission_handler
 from lego.apps.polls.models import Poll
-from lego.apps.polls.serializers import DetailedPollSerializer
+from lego.apps.polls.serializers import (
+    DetailedPollSerializer,
+    HiddenResultsDetailedPollSerializer,
+)
 from lego.apps.users.models import User
 
 
@@ -94,11 +97,17 @@ class FrontpageViewSet(viewsets.ViewSet):
         events = FrontpageEventSerializer(
             queryset_events, context=get_serializer_context(), many=True
         ).data
-        poll = (
-            DetailedPollSerializer(queryset_poll, context=get_serializer_context()).data
-            if queryset_poll
-            else None
-        )
+        poll = None
+        if queryset_poll:
+            if queryset_poll.results_hidden:
+                poll = HiddenResultsDetailedPollSerializer(
+                    queryset_poll, context=get_serializer_context()
+                ).data
+            else:
+                poll = DetailedPollSerializer(
+                    queryset_poll, context=get_serializer_context()
+                ).data
+
         ret = {"articles": articles, "events": events, "poll": poll}
 
         return Response(ret)
```
</p>
</details>

LarsSelbekk:
Okay, I applied your patch, thanks for that. I set you as the author of that commit, I can change it if you want to. After a long time searching I finally fixed the pipeline failure. Turns out someone else had the same problem in commit f943daa0.

LarsSelbekk:
Frontend PR is webkom/lego-webapp#2190.

LudvigHz (approved):

PR #2287, opened 2021-08-23
Fix normalization issue in PG search queries

LudvigHz:
For autocomplete we create the search vector manually. This created
problems with queries containing reserved symbols used by PG when
creating the vector. This removes all reserved symbols from the query.

See sentry for issues this fixes.
As an example, if someone searches for "Noe & kult", the search will crash because the query sent to the search is `noe & & & kult` which makes no sense.

### DIFF
--- a/lego/apps/search/index.py
+++ b/lego/apps/search/index.py
@@ -108,6 +108,17 @@ class SearchIndex:
         """
         return None
 
+    def clean_query(self, query):
+        """
+        Clean search query to prepare for pg search.
+        Removes characters like &, | and other chars used in pg full text search.
+        """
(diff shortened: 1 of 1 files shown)


lego/apps/search/index.py
```diff
@@ -108,6 +108,17 @@ def get_autocomplete(self, instance):
         """
         return None
 
+    def clean_query(self, query):
+        """
+        Clean search query to prepare for pg search.
+        Removes characters like &, | and other chars used in pg full text search.
+        """
+        chars = ["&", "*", ":", "|", "<>", "!"]
```
amalieeh, line 116: How come "<>" are together?

PR #2330, opened 2021-09-21
Fix the order of the commands in READE.md

juniwbjerde:
The order of the commands presented in the "initial setup" step in README.md is wrong. One has to run docker-compose up -d before running python manage.py initialize_development


SmithPeder (approved):
Nice

LudvigHz (changes requested):
Maybe remove the command from the block below, or split this into a "python" part, "docker" part and a "set up" part

SmithPeder:
Ups, I think I enabled auto-merge :O

PR #2331, opened 2021-09-21
Allow allergies field to be up to 500 chars

### LINKED ISSUE #2310: Allow more characters for the allergies field
Like it is now, one can not have more than 100 characters in the allergies field for a user (the field in question is found in settings on the profile for a user on abakus.no). It turns out that some users need more space than this to be able to fit all their allergies, so we need to expand the limitation on this field. 

Tip: The validation for this field is done backend. A good start could be to search for "allergies" and see what you find then:)

EikHR:
Closes #2310


SmithPeder (approved):
LGTM

LudvigHz (approved):
Lgtm :shipit:

PR #2338, opened 2021-10-05
Replace hard-coded http status code integers with constants from rest_framework

### LINKED ISSUE #2329: Use status from rest_framework to evaluate tests
A lot of the tests in backend use response-codes (200 ok, 403 forbidden etc.) in their assertionstests. Some of them are importing constants from the `status` class within the `rest_framework` library (`status.HTTP_200_OK`, and `status.HTTP_403_FORBIDDEN`), while others are hard-coded integers. The integers should ideally be replaced by the `rest_framework` constants to increase readability.

Following regex can be used to find all occurrences: `, [0-9]{3}\)`

eikhr:
Closes #2329

Replaces all hard-coded status codes in tests with constants imported from the `status` class within the `rest_framework` library.

### DIFF
--- a/lego/api/test_urls.py
+++ b/lego/api/test_urls.py
@@ -1,4 +1,5 @@
 from django.conf import settings
+from rest_framework import status
 
 from lego.utils.test_utils import BaseTestCase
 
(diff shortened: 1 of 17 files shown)


PeterJFB (approved):
lgtm👏

PR #2359, opened 2021-10-20
Add group view restrictions for events

iselinee:
Add backend for restricted view permissions on events.

### DIFF
--- a/lego/apps/events/tests/test_events_api.py
+++ b/lego/apps/events/tests/test_events_api.py
@@ -34,7 +34,7 @@ _test_event_data = [
         "startTime": "2011-09-01T13:20:30Z",
         "endTime": "2012-09-01T13:20:30Z",
         "mergeTime": "2012-01-01T13:20:30Z",
-        "isAbakomOnly": False,
+        "canViewGroups": [],
         "pools": [
             {
                 "name": "Initial Pool",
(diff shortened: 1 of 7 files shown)


LudvigHz (changes requested):
Looks good!
Before we can merge this, we need to migrate all old events, so that they still work. We will probably need to create a custom migration to add `Abakom` to `can_view_groups` for all events that have `abakom_only=True`.

lego/apps/events/tests/test_events_api.py
```diff
@@ -310,28 +334,6 @@ def test_creator_see_registrations(self):
         for pool in event_response.json()["pools"]:
             self.assertIsNotNone(pool.get("registrations", None))
 
-    def test_without_auth_permission_abakom_only(self):
```
erlingfn, line 313: Did we have two copies of the same test?

erlingfn (approved):
Looks good! Great to get this fixed

eikhr (approved):

lego/apps/events/views.py
```diff
@@ -307,6 +308,10 @@ def create(self, request, *args, **kwargs):
         serializer.is_valid(raise_exception=True)
         event_id = self.kwargs.get("event_pk", None)
         event = Event.objects.get(id=event_id)
+
+        if not get_permission_handler(Event).has_perm(request.user, VIEW, obj=event):
+            raise PermissionDenied()
```
ivarnakken, line 313: Good catch 💯

ivarnakken (approved):
Let's finally get this thing in ... 👴🏼

lego/apps/events/tests/test_events_api.py
LudvigHz: Looks like it yeah

LudvigHz (approved):

PR #2369, opened 2021-10-31
(WIP) Add view for deleting user

theodorsm:
Linked to https://github.com/webkom/lego-webapp/pull/2304

Which permissions should be needed to delete users? Only the user itself, sudo or AbakusGroup?


lego/apps/users/views/users.py
```diff
@@ -173,3 +173,11 @@ def change_grade(self, request, *args, **kwargs):
                 newGrade.add_user(user)
 
         return Response(MeSerializer(user).data)
+
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+        self.perform_destroy(instance)
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    def perform_destroy(self, instance):
```
PeterJFB, line 182: I'm no django expert, asking out of curiosity:P Some of the logic here seems similar to 

https://github.com/webkom/lego/blob/dbe2dcc3466be7b0c934ff5d0d5c5a403261b4c5/lego/apps/users/views/user_delete.py#L11-L17

Do we use these methods for different purposes?:)

theodorsm: The destroy and perform_destory is triggered by an HTTP DELETE request from a user with correct permissions. This is code from an implementation I changed and is used now. But could be useful for deleting users without shelling. I will look into removing this code or keeping it for admin deletion.

theodorsm: Changed the destroy to do nothing in latest commit, to make sure only the correct endpoint is used for deleting users.

lego/apps/users/views/user_delete.py
```diff
@@ -0,0 +1,17 @@
+from rest_framework import permissions, status, viewsets
```
LudvigHz, line 1: I feel like this is small enough to be in `views/users` but :shrug:

LudvigHz (approved):

lego/apps/users/views/user_delete.py
theodorsm: Just followed the convention from ChangePasswordViewSet, which is also small, but is separated out in its own view file.

PeterJFB (approved):
lgtm^^

PR #2372, opened 2021-11-07
Add active tag for groups

SmithPeder:
Adds an active field to the groups. Will initially be used to show active and inactive interestgroups.


ivarnakken (approved):
Love it! 💯

PR #2439, opened 2021-12-19
Use first name instead of full name on all mails

ivarnakken:
It's been bugging me a lot lately that the mails we send out use people's full names, and not their first. I believe a first name is more personal, and feels less "automated", in an otherwise quite automated mail ... 😄 In addition, my inner ocd kicked in once I realized we were inconsistently utilizing commas and exclamation marks 😢


SmithPeder (approved):
LGTM

PR #2479, opened 2022-01-16
Add LQ images for progressive image loading

LudvigHz:
~This field can be used to automatically create a placeholder image beside the full image, that can be used as a placeholder.~

Adds plenty of `*_placeholder` images used for progressive image loading.

### DIFF
--- a/lego/apps/articles/serializers.py
+++ b/lego/apps/articles/serializers.py
@@ -17,6 +17,9 @@ class DetailedArticleSerializer(TagSerializerMixin, BasisModelSerializer):
     author = PublicUserSerializer(read_only=True, source="created_by")
     comments = CommentSerializer(read_only=True, many=True)
     cover = ImageField(required=False, options={"height": 500})
+    cover_placeholder = ImageField(
+        source="cover", required=False, options={"height": 50, "filters": ["blur(20)"]}
+    )
     content_target = CharField(read_only=True)
     content = ContentSerializerField(source="text")
     reactions_grouped = serializers.SerializerMethodField()
@@ -31,6 +34,7 @@ class DetailedArticleSerializer(TagSerializerMixin, BasisModelSerializer):
             "id",
             "title",
             "cover",
+            "cover_placeholder",
             "author",
             "description",
             "comments",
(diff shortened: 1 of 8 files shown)


LudvigHz:
### UPDATE:
Now with less content shift and more/better loading. I have gone through and most places now should not shift content at all when partially loading.

Also: there were some really wonky stuff with the articles page that I hadn't noticed before, like misaligned items and weird borders. Should be good now :smile: 

https://user-images.githubusercontent.com/42850232/153245458-51543555-5fa4-4401-ad07-b7b5fa578336.mp4

https://user-images.githubusercontent.com/42850232/153245591-9514ea45-e486-456f-ab67-6354ca161527.mp4

lego/apps/articles/serializers.py
```diff
@@ -75,6 +79,9 @@ class Meta:
 class PublicArticleSerializer(TagSerializerMixin, BasisModelSerializer):
 
     cover = ImageField(required=False, options={"height": 300})
+    cover_placeholder = ImageField(
+        source="cover", required=False, options={"height": 50, "filters": ["blur(20)"]}
```
SmithPeder, line 83: this should be 30 ey?

SmithPeder (changes requested):
If i get the schema correctly, my comments make sense?

lego/apps/companies/serializers.py
```diff
@@ -86,6 +86,9 @@ def create(self, validated_data):
 
 class CompanyListSerializer(BasisModelSerializer):
     logo = ImageField(required=False, options={"height": 500})
+    logo_placeholder = ImageField(
+        source="logo", required=False, options={"height": 500, "filters": ["blur(20)"]}
```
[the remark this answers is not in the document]
LudvigHz: oh damn. yup

lego/apps/articles/serializers.py
LudvigHz: yup

LudvigHz:
> Shouldn't the SearchArticleSerializer class also use a cover placeholder?

No. This is intentional, as this serializer is only used when searching, where we don't use the cover. If you think it should, sure. But I don't think it needs to be there. (We don't use the cover on lego-webapp at least).

SmithPeder (approved):

PR #2536, opened 2022-02-23
Allow unauthenticated users list & view permission on companies

LudvigHz:
This first of all allows proper use of the companyinterest form, where a
user can select their company when filling in the form, without the need
for a user. It also opens up for listing all companies, as well as
viewing their details, without authentication.

I don't know what the rationale was to not allow this before, and use the `isAuthenticated`
permission class on the company viewset, but I at least see no reason this should not be possible.


SmithPeder:
> It also opens up for listing all companies, as well as viewing their details, without authentication. I don't know what the rationale was to not allow this before.

We've had this discussion in both Webkom and itDagene. The issue has always been
"Do we expose our list of companies to other companies"

Both times I've had mixed feelings, but I have little against this now and agree that we can enable it. The amount of companies at Abakus is at the scale that being in our BDB is no indicator of anything. It says nothing about our involvement whatsoever.

SmithPeder (approved):
Remember to 
- create migrations 
- fix linting
- fix tests
 😙 
 😙 
 😙 
 😙

juniwbjerde (approved):
LGTM!

LudvigHz:
> Both times I've had mixed feelings, but I have little against this now and agree that we can enable it. The amount of companies at Abakus is at the scale that being in our BDB is no indicator of anything. It says nothing about our involvement whatsoever.

Exactly. And I figure that when you can see both joblistings and company events without logging in, that says a lot more about the companies invested in Abakus than a simple list of companies related to Abakus in some way. Exposing this list yields absolutely no more info than what's already available.

LudvigHz:
> create migrations
fix linting
fix tests

I was a little quick on the trigger here I see

SmithPeder (approved):
LGTM 👍🏻 
Kinda hard to diff with Githubs shitty "diff-over-diff-files" view, but it looked okey

PR #2554, opened 2022-03-08
Disallow users from approving their own quotes

ivarnakken:
This was discussed in today's (08.03.2022) meeting. 

This has to be done in the backend, as the creator of quotes shall remain "anonymous" to the client. 
I thought doing the check on `approve` was a better approach than filtering the quotes from the list because one _might_ get confused as to why their quote is not showing up after creating one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ If you want to check your test coverage, you can do the following
 
 ```bash
 # Install the packages needed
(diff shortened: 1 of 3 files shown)


theodorsm (approved):
LGTM :)

SmithPeder (approved):
LGTM

PR #2560, opened 2022-03-10
fix sending of mails to too many people

### LINKED ISSUE #1299: RestrictedMail notification
> Restricted mail is used when sending mails to multiple users at once by selecting users/events/meetings, and then send the email to <restricted@abkaus.no> together with the token.

The `restricted mail sent` should be sent to the proper email, not the `user.email` field. The address `user.email_address` should be used instead.

If the `from_address` is not the same as the `user.email_address`, both should receive the mail.

https://github.com/webkom/lego/blob/ccab14fbee223f16842ace6ca2ba0c2f3ac3ac86/lego/apps/restricted/notifications.py#L9

Arashfa0301:
- Now mails won't be sent if the number of users receiving is more than 500. 
- also fixed #1299

### DIFF
--- a/lego/apps/restricted/serializers.py
+++ b/lego/apps/restricted/serializers.py
@@ -1,3 +1,7 @@
+from functools import reduce
+
+from rest_framework import exceptions
+
 from lego.apps.events.fields import PublicEventListField
 from lego.apps.meetings.fields import MeetingListField
 from lego.apps.restricted.models import RestrictedMail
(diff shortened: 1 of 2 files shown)


lego/apps/restricted/serializers.py
```diff
@@ -28,6 +30,17 @@ class Meta(RestrictedMailListSerializer.Meta):
...
         )
 
+    def create(self, validated_data):
+        groups = validated_data["groups"]
+        events = validated_data["events"]
+        num = reduce((lambda a, b: a + b.number_of_users), groups, 0)
+        num += reduce((lambda a, b: a + b.registration_count), events, 0)
+        if num > 500:
+            raise exceptions.ValidationError(
+                "The number of students in selected groups/arrangements exceed the permitted amount"
```
ivarnakken, line 40: Do you mean "arrangements" as in "events"?

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
```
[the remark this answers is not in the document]
Arashfa0301: Had a problem with my vscode where it ran prettier every time i saved. All good now 👍

lego/apps/restricted/serializers.py
Arashfa0301: Right on 😊

juniwbjerde (approved):
LGTM!

PR #2580, opened 2022-03-15
Inactive counter is reset on login

theodorsm:
The counter is never reset as of now. If a user receives enough mails and logs in again and later is inactive, a mail may not be sent before deletion.


Arashfa0301 (approved):
LGTM!!

eikhr (approved):

PR #2605, opened 2022-03-24
feat: add meeting search index

AdrianAndersen:
Add meeting search index for meetings.

This makes it possible to search for meeting titles, meeting descriptions and meeting reports.

Relevant frontend PR: webkom/lego-webapp#2584

### DIFF
--- a/lego/apps/meetings/search_indexes.py
+++ b/lego/apps/meetings/search_indexes.py
@@ -1,0 +1,32 @@
+from lego.apps.meetings.models import Meeting
+from lego.apps.meetings.serializers import MeetingSearchSerializer
+from lego.apps.search import register
+from lego.apps.search.index import SearchIndex
+
+
+class MeetingModelIndex(SearchIndex):
+
+    queryset = Meeting.objects.all()
+    serializer_class = MeetingSearchSerializer
+    result_fields = (
+        "title",
+        "description",
+        "report",
+        "start_time",
+    )
+    autocomplete_result_fields = ("title", "start_time")
+
+    autocomplete_fields = ("title", "description")
+    search_fields = ("title", "description", "report")
+
+    def get_autocomplete(self, instance):
+        return instance.title
+
+    def search(self, query):
+        return super().search(query).order_by("-start_time")
+
+    def autocomplete(self, query):
+        return super().autocomplete(query).order_by("-start_time")
+
+
+register(MeetingModelIndex)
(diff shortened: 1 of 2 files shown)


lego/apps/meetings/search_indexes.py
```diff
@@ -0,0 +1,34 @@
...
+    serializer_class = MeetingSearchSerializer
+    result_fields = (
+        "title",
+        "description",
+        "report",
+        "start_time",
+    )
+    autocomplete_result_fields = ("title", "start_time")
+
+    autocomplete_fields = ("title", "description", "report")
```
LudvigHz, line 21: Remove `"report"` here. It doesn't make sense to autocomplete based on the content of the meeting report.

LudvigHz (changes requested):
Awesome!

lego/apps/meetings/search_indexes.py
AdrianAndersen: The meeting report is the actual content of the meeting, while the description is the short message in calendars. I think it makes sense to be able to search for something from the report. (aka. the meeting notes). At least in our case, it is not very useful to only search for the meeting title. What do you think?

LudvigHz: Yes, that's why it makes sense to have it in `search_fields`, which is exactly what this field is for (as you have done). But not in `autocomplete_fields`, as that would just be confusing and not make sense.

AdrianAndersen: Hm, that is weird. In my testing, I have to include "report" in the autocomplete_fields as well to make it actually show up in the results. 🤔

LudvigHz: So, the reason for this is kinda weird/stupid. In lego-webapp, when you search in the top bar, and wait for results to show up as you type, we use the `/autocomplete` endpoint, and the full search is only done once you press Enter.

This is kinda strange and confusing, and we have discussed before what the best behavior is here, but haven't changed it, or made it more clear that the search-as-you-type results aren't full search. The behavior I describe we want for meetings, is also how it works for events, pages, articles etc. If you search for page content in lego-webapp now, you'll notice you won't get any results, but when you press enter, you will see the expected results.

LudvigHz: It does say "Trykk enter for fullstendig søk at the bottom of the results" but it's not clear enough what the behavior is I guess :upside_down_face:.

AdrianAndersen: Ah yes, I see. Then it makes sense, I agree.

SmithPeder (approved):

LudvigHz (approved):

PR #2619, opened 2022-03-30
Don't allow successive creation of registrations

LudvigHz:
This fixes another race condition (which is not really a race condition, but just bad logic). You
could see this if you register from two devices, and click "register" a second later than the first
one. This would cause the first request to successfully register, but the second request would be
oblivious to this, fail and set the status to `FAILURE_REGISTER`. This PR should fix this, and also
will catch the case where the user is already registered earlier, without having to register and run
the registration task. This might make it harder to "fix" a stuck registrations by submitting
another one, although this should :tm: not happen, and this PR should fix issues that cause just
this case.

If you send two requests to create a registration, such that the first
request finishes processing, and register() finishes successfully, the
second request will set the registration status to PENDING_REGISTER,
the registration task will be processed and fail, and since the status
has been overwritten, the status will now be set to FAILURE_REGISTER,
even though the registration is valid and successfull (from the first
request).

### DIFF
--- a/lego/apps/events/tests/test_events_api.py
+++ b/lego/apps/events/tests/test_events_api.py
@@ -992,6 +992,28 @@ class RegistrationsTransactionTestCase(BaseAPITransactionTestCase):
         )
         self.assertEqual(res.json()["status"], constants.FAILURE_REGISTER)
 
+    def test_successive_registrations(self, *args):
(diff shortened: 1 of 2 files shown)


erlingfn (approved):
Looks good to me!

PR #2620, opened 2022-03-30
Add two year masters option to student confirmation

LudvigHz:
Allows users to select whether they are enrolled in two year masters programme, so they are put in the right grade when registering.


SmithPeder (approved):
Very smart feature, logic and tests looks good to me!

ivarnakken (approved):
This will be so useful in august! 😍 Love the tests as well!

PR #2624, opened 2022-03-31
Select for update initial registration

hanswilw:
Blind fix with zero testing. 
My _guess_ is that simultaneous registration attempts ends up changing it back from `SUCCESS_REGISTER` to `PENDING_REGISTER`, then becoming `FAILURE_REGISTER`. The create api did not use the select for update lock atleast :)

I slapped on an `of=("self",)` in case it locks the event and user, dont remember if it does that for primary keys – but still ok to explicitly only lock the registration.

### DIFF
--- a/lego/apps/events/views.py
+++ b/lego/apps/events/views.py
@@ -312,11 +312,15 @@ class RegistrationViewSet(
         current_user = request.user
 
         with transaction.atomic():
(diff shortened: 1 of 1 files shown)


hanswilw:
Closing in favor of https://github.com/webkom/lego/pull/2619

PR #2642, opened 2022-04-08
Retrieve surveys in reverse order

SmithPeder:
I don't know if you agree that this makes sence, but it's really annoying that i always get the survey from 2018 at the top, and the one from 2022 at the bottom...﻿


LudvigHz (approved):
This has really annoyed me as well :100: :shipit: :rocket:

PR #2643, opened 2022-04-10
Fix ContentField img sanitizer

theodorsm:
In some events the text contains broken images, instead of raising errors and not being able to update the text for the event or remove it, it's better to just log a warning. 

In the front end the broken image is shown like this:
image

### DIFF
--- a/lego/apps/content/fields.py
+++ b/lego/apps/content/fields.py
@@ -3,11 +3,14 @@ from django.db.models import TextField
 from rest_framework import serializers
 
 from bs4 import BeautifulSoup
+from structlog import get_logger
 
 from lego.apps.content.utils import sanitize_html
 from lego.apps.files.constants import IMAGE, READY
 from lego.apps.files.models import File
 from lego.apps.files.thumbor import generate_url
+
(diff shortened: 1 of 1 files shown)


LudvigHz:
Hmmm, the thing is that we don't want the event to be saved and we want to throw an error when someone tries to do this. But it might be better to just move the validation up somewhere, so that it's possible to run `event.save()` without it breaking everything and still have the current behavior when a user tries to do the same

theodorsm:
> Hmmm, the thing is that we don't want the event to be saved and we want to throw an error when someone tries to do this. But it might be better to just move the validation up somewhere, so that it's possible to run `event.save()` without it breaking everything and still have the current behavior when a user tries to do the same

I agree that's a better solution, but I suggest we deploy this as a "hotfix" and revert it back to raising errors after the first round of deleting users. I do not have time to work out a propper fix before tomorrow morning when it all goes down.

LudvigHz (approved):

theodorsm:
Closed due to fixing the bad events throwing errors

PR #2773, opened 2022-07-08
Update README regarding coverage and implement Codecov

ivarnakken:
* Add a coverage badge from Codecov to the top, which I hope will inspire people (myself included) to test more of their code
* Improve docs on how to obtain a coverage report

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -114,11 +114,16 @@ steps:
       event: push
     environment:
       LANG: C.UTF-8
+      CACHE: cache
+      CODECOV_TOKEN:
+        from_secret: codecov_token
     depends_on:
       - tests
     commands:
       - pip install tox
       - tox -e coverage
+      # Upload coverage to Codecov
+      - codecov -t ${CODECOV_TOKEN} -f "coverage.xml"
 
   - name: docker
     image: plugins/docker
@@ -190,4 +195,4 @@ services:
 
 ---
 kind: signature
-hmac: 38a22c24e963a667013c35a2113df448f2c0b534d098b8cd3c5424bedefedec0
+hmac: 51c80c10b5360d90aa1e6c0361519fb3b7485f2782830e492d3580774625483a
--- a/.drone.yml
+++ b/.drone.yml
@@ -10,7 +10,7 @@ trigger:
 
 steps:
   - name: setup
-    image: python:3.9
+    image: abakus/lego-testbase:python3.9
     when:
       event: push
     environment:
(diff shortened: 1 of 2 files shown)


eikhr (approved):
Cool! LGTM

SmithPeder (commented):
I like the result!! Not a huge fan of implementing `curl` + `chmod` in the middle of a pipeline through. Is there no official python codecov images out there? 

An alternative would be to install codecov into [lego-testbase](https://github.com/webkom/lego-testbase) image alongside `tox`. But as of writing this I see that @warlo actually updated lego-testbase to `python 3.9` in [this commit](https://github.com/webkom/lego-testbase/commit/fe2424c1bc8f2d441133e48a7624b019e5d357b4), but then continued to remove lego-testbase in [this commit](https://github.com/webkom/lego/commit/f62a0259c9b582ef32bcac9d02de8028e9af5864) 😨 . I was not aware of this change, and I do condemn it 😄

ivarnakken:
> Is there no official python codecov images out there?

[This is how](https://docs.codecov.com/docs/codecov-uploader#using-the-uploader) they want you to do it, and there is no official image. However, I did try out https://github.com/robertstettner/drone-codecov, but it never worked :( 

> But as of writing this I see that @warlo actually updated lego-testbase to python 3.9 in [this commit](https://github.com/webkom/lego-testbase/commit/fe2424c1bc8f2d441133e48a7624b019e5d357b4), but then continued to remove lego-testbase in [this commit](https://github.com/webkom/lego/commit/f62a0259c9b582ef32bcac9d02de8028e9af5864) :fearful:

Yea I also thought that was strange. @warlo, care to explain?

warlo:
I think we did that because lego-testbase had to be built and pushed to the registry and I didnt have access IIRC. Then since it only installed tox @LudvigHz (I think) and I agreed we might as well just use default build and install it manually. You can revert that commit and use the lego image if you prefer :)

LudvigHz:
Yeah, I remember it the same. `lego-testbase` is kinda pointless, and it's a _bit_ annoying having a separate docker image for just a single pip dependency and a binary or two 😄 

I think the curl is fine tbh. The only thing is that it would be nice to cache it, but 🤷 (I'm guessing it's tiny anyhow). The current caching situation needs to be fixed anyways due to the ephemeral drone runners.

ivarnakken:
I reverted back to employ the `lego-testbase` image, and [added the Codecov installation to it](https://github.com/webkom/lego-testbase/pull/3/).

The fact that Drone did not automatically pull the newer image version since it was not tagged with `:latest` was truly a defeat and costed me a great deal of pain ... 😣

PR #2801, opened 2022-07-26
Support Django 4

ivarnakken:
## The future is now

I've migrated what seems to be _necessary_ in order to support Django 4. But, the new version brings a bunch of new interesting features that I would like to have a closer look at (potentially in a new PR). Nevertheless, this should be "ready" to review.

`djangorestframework-jwt` is replaced with `drf-jwt`, which is simply an "updated" fork. However, I think `djangorestframework-simplejwt` has some cool features ([and the author of `djangorestframework-jwt` agrees](https://github.com/jpadilla/django-rest-framework-jwt/issues/484)), so I naturally begun to migrate to it but realized in the midst that it was kinda out of scope of what I'm trying to achieve with this PR, so `drf-jwt` will have to do for now - but will be replaced soon! Also, it's nice to diversify breaking changes into multiple PRs to minimize risk 😅


erlingfn (commented):
The code looks good, however there seems to be a few errors with authentication still in

LudvigHz (commented):
Haven't looked through the change notes to check if anything is missing. But just from looking at the changes here everything looks good!

`missing-migrations` runs fine locally, so not quite sure why it fails in CI though :thinking:

norbye (approved):
Looks good:) weird with the missing migrations tho, struggeling get it to not work locally

ivarnakken:
> [...], however there seems to be a few errors with authentication still in

@erlingfn. The auth problems are fixed in https://github.com/webkom/lego-webapp/pull/2924

ivarnakken:
From [this related Stack Overflow question](https://stackoverflow.com/questions/52745693/django-db-utils-programmingerror-relation-django-content-type-does-not-exist) I'm suspecting that `missing-migrations` is failing in CI because the db is not fully ready during the first migration, yet Drone is trying to access the db. Does this sound like a possible reason?

I recommend running Drone locally for a better traceback of the error.

ivarnakken:
I've done some extensive testing with this in staging, and everything seems to work as intended :tm:. Would appreciate if others had a go with is as well :heart:

PR #2829, opened 2022-08-08
Change wording of student confirmation email

juniwbjerde:
As NTNU recently blocked our mails because the wording of this confirmation email made it sound like phishing, we figured it was time to change it a bit:)


ivarnakken (approved):
That's better 💯 

Just fix up the flake8 and black errors and we're good!

PR #2830, opened 2022-08-08
Fix make commands by changing python image

ivarnakken:
make commands were not working, due to `git` being needed by `isort` and yet not in the slim python image. Not using the slim version here does not hurt the performance notably.


juniwbjerde (approved):
Good catch!

PR #2835, opened 2022-08-11
Notify users of unanswered meeting invitations 

ivarnakken:
I personally think we should notify users of when they have misbehaved and not answered their meeting invitations.

I also took the liberty to change the color of the "accept" button on the mails.


lego/apps/meetings/tasks.py
```diff
@@ -0,0 +1,40 @@
...
+    self.setup_logger(logger_context)
+
+    meetings: list[Meeting] = Meeting.objects.filter(
+        start_time__gt=timezone.now()
+    ).prefetch_related("registrations")
+
+    for meeting in meetings:
+        # Only send reminder after 2 days
+        if meeting.created_at > timezone.now() - timedelta(days=2):
+            continue
```
LudvigHz, line 27: You can do this check in the query :) Just like with the start time check.

e.g.
```python
Meeting.objects.filter(created_at__lt=timezone.now() - timedelta(days=2))
```

LudvigHz (commented):
Does this mean the user will get a notification _every day_ until they answer the meeting invite?
Maybe that's a bit much?

lego/apps/meetings/tasks.py
ivarnakken: Ah yes of course, that's cleaner! I'm assuming you mean `created_at__gt` instead of `created_at__lt` though 👍🏼

ivarnakken:
> Does this mean the user will get a notification every day until they answer the meeting invite?

Yea, and I suppose it might be a bit much. I can make the cronjob run every other day instead, e.g. on "even" days: `0 0 2-30/2 * *`.  Would that be better? @LudvigHz

lego/apps/meetings/tasks.py
norbye: Shouldn't it be lt? A timestamp from last year would be less(smaller number) than a timestamp from two days ago, while todays timestamp would be greater than

ivarnakken: Oh shoot you're right, I was simply looking at the if-statement that I so carefully constructed yesterday, but when filtering the query, it's the other way around. Sorry!

lego/apps/meetings/tasks.py
```diff
@@ -0,0 +1,41 @@
...
+
+@celery_app.task(serializer="json", bind=True, base=AbakusTask)
+def notify_user_of_unanswered_meeting_invitation(self, logger_context=None):
+    self.setup_logger(logger_context)
+
+    meetings: list[Meeting] = Meeting.objects.filter(
+        start_time__gt=timezone.now(),
+        # Only send reminder 2 days after creation
+        created_at__lt=timezone.now() - timedelta(days=2),
+    )
```
norbye, line 24: Could it be an idea to limit this to a timeframe as well? So that you will only be notified if the meeting is in the next x days?

For context I just thought about inviting to a series of meetings with regular frequency - and that it could be nice to just create all the meetings in the beginning of the semester and for people to answer it 1-2 weeks in advance when they know what they'll be up to

ivarnakken: Yep, that can also be done😄 How long before do you think is suitable? Two days?

norbye: Whopsies, forgot to reply to this one.. How about notifying the recipients if they haven't answered, starting 7 days in advance, and then notifying them every other day until the day before? so 7, 5, 3, 1 days before the date of the meeting.

ivarnakken: Sure! I'll see what I can do!

ivarnakken: It now only sends reminder when it's less than 7 days and either 7, 5, 3 or 1 day till start.

norbye (approved):
Noice :heart_eyes:

LudvigHz (approved):

PR #2840, opened 2022-08-12
Add a few additional categories for flatpages

erlingfn:
This is a simple prerequisite for webkom/lego-webapp#2708, just putting the new categories as options in the backend


ivarnakken (approved):
Nice! 💯 You'll need to make migrations for this change though, and then we're all good!

SmithPeder (approved):

LudvigHz (approved):

eikhr (approved):

ivarnakken:
Yea, and btw, you'll have to run black on the auto generated migration files as well 😅

erlingfn:
> And, amend your commits pls grimacing
Why? It will be squashed when it is merged, and force pushing after people have reviewed can be quite annoying. It does matter in this case, but why take care to make the commit log clean when it will be squashed before merging?

ivarnakken:
> It will be squashed when it is merged

Indeed. It will be squashed because I made it so. It's _ok_ in this case, but you don't always want to squash all the commits in a PR, and in those cases it's simply good practice to amend. So, it was said mostly said for good measure later I guess

erlingfn:
> Indeed. It will be squashed because I made it so. It's ok in this case, but you don't always want to squash all the commits in a PR, and in those cases it's simply good practice to amend. So, it was said mostly said for good measure later I guess

This is a discussion we can have in-office, but our practice of force pushing while people are reviewing might lead to some issues, since if one force pushes after someone have reviewed, it might be very hard to tell what is changed since last review.

ivarnakken:
> it might be very hard to tell what is changed since last review.

You can still compare the commits though, but I see your point 👍🏼

PR #2859, opened 2022-08-21
Fix method to retrieve jwt token

LudvigHz:
This fixes the WS issues in prod. Also it enables some logging, since the server was throwing a
bunch of errors and crashing, but the logging config meant none of them got printed anywhere.

We recently upgraded to a fork of the jwt library we are using. This had
changed the method name used for retrieving the jwt token, breaking the
jwt middleware in the websocket server.


ivarnakken (approved):
Nice! 💯 Go go go

PR #2879, opened 2022-08-25
Make sure meeting object permissions are checked in search

LudvigHz:
Object permissions were not properly checked for meeting objects when searching. Meaning some
information about meetings were leaked when searching.


erlingfn (approved):
Looks good, just need to fix the problem with CI

ivarnakken (approved):

PR #2884, opened 2022-08-26
Filter on meeting `invitations` instead of `invited_users`

ivarnakken:
`invited_users` would give us the invited users, but once we used their `invitations__status` we would receive _all_ of the user's invitations (and not just for the specific meeting).


SmithPeder (changes requested):
'lego.apps.users.models.User' imported but unused

LudvigHz (approved):

PR #2886, opened 2022-08-28
Return group type from AbakusGroup field

ivarnakken:
Group `type` was "used" by the webapp on its event pages, but was for obvious reasons `undefined`.

image


LudvigHz (approved):
LGTM :shipit:

norbye (approved):

PR #2902, opened 2022-09-01
Add unique_together constraint on photo consents

LudvigHz:
Somehow we ended up with some PhotoConsent objects with the exact same
values.

To explain some more how this happened: using `get_or_create` and `update_or_create` requires that we have set proper database constraints, otherwise a race condition might cause multiple rows to be inserted, thus ending up where we are now.

See the BIG WARNING box in the docs for these methods: https://docs.djangoproject.com/en/4.1/ref/models/querysets/#get-or-create


ivarnakken (approved):

eikhr (approved):
nice

lego/apps/users/views/users.py
```diff
@@ -184,19 +184,15 @@ def destroy(self, request, *args, **kwargs):
...
             raise PermissionDenied(detail="Cannot update other user's consent")
 
-        PhotoConsent.objects.update_or_create(
-            user=user,
-            year=serializer.validated_data["year"],
-            semester=serializer.validated_data["semester"],
-            domain=serializer.validated_data["domain"],
-            defaults={"is_consenting": serializer.validated_data["is_consenting"]},
-        )
+        serializer = self.get_serializer(data={"user": user, **request.data})
```
ivarnakken, line 192: Any reason for changing from `data=request.data`? Is the user not in the data?

ivarnakken (approved):
Looks good still 💯 The init process felt kinda overkill and weird to me, but it'll do

lego/apps/users/views/users.py
LudvigHz: No, as you can see, the user is being retrieved in the `get_object` method call further up. If you look at the old code, it was this user instance that was being user elsewhere as well.

PR #2915, opened 2022-09-06
Optimize photo_consents queries on administrate endpoint

### LINKED ISSUE #2909: administrate endpoint is insanely slow
See django debug toolbar when making a request.
Due to the way that PhotoConsents are fetched for the users, we make around 70 SQL requests for 19 users. This grows for the number of users as well.

The solution to this is to make a better query in the view, either annotating the query, or prefetching and setting the prefetched query as an attribute on the queryset. See django `Prefetch()` objects.

At the same time, we should also limit the photoConsent objects that are fetched for each user to just those that are relevant for the specific event, and maybe those on the registration object.

LudvigHz:
Fixes #2909

The administrate endpoint was super slow, since we did several queries per user registered to the
event. Each user registered would produce 3 extra queries to the DB. So for an event with just 9
registrations, we previously had 75 queries. With this, we are down to 20.

For context, for an event with 100 registered users, this would be ~350 queries.

### DIFF
--- a/lego/apps/events/fields.py
+++ b/lego/apps/events/fields.py
@@ -1,8 +1,9 @@
 from rest_framework import serializers
 from rest_framework.exceptions import PermissionDenied
 
-from lego.apps.events.models import Event
+from lego.apps.events.models import Event, Registration
 from lego.apps.permissions.constants import EDIT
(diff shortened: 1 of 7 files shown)


eikhr (approved):
lgtm

PR #2945, opened 2022-09-19
Update README.md

jespeste:
Changed then to than, as that is the correct use of the english word than


erlingfn (approved):
LGTM

ivarnakken (approved):
Thank you!

PR #2962, opened 2022-09-28
Add created_after filtering for joblistings

LudvigHz:
Also fixes thumbnails on joblistings to not be cropped. The new
parameters mean that images are filled with white color instead of
cropped in the other direction.

This is needed for the new slack notifications for new joblistings.


mrfarstad:
LGTM 🚀🎸🕺

ivarnakken (approved):
💯

PR #2963, opened 2022-09-29
Make sure possible_pools are distinct

LudvigHz:
Due to the way django querysets filter on many-to-many relationships, it
will return duplicate values for _each of the relationships_ where the
filter condition is true. Slapping a `.distinct()` fixes this.

This fixes an issue where the `spots_left` count is wrong for certain users
if the pool contains several permission groups overlapping with the users' groups.
I.e. if you are in 5.th grade, and the pool contains **both** `Abakus` and `5. klasse`,
then the `spots_left` attribute would count the spots left for _each_ of these groups, even though
they are both in just _one_ pool.

See https://stackoverflow.com/a/30262473 for more info on why this happens.

### DIFF
--- a/lego/apps/events/models.py
+++ b/lego/apps/events/models.py
@@ -226,7 +226,7 @@ class Event(Content, BasisModel, ObjectPermissionsModel):
             is_admitted = self.is_admitted(user)
         if is_admitted:
             return []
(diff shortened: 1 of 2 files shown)


LudvigHz:
We should also make sure that this issue is not present other places.
I did a _very quick_ look over, and couldn't find any more obvious cases, but we should do a more thorough check.

PR #2979, opened 2022-10-05
Weekly email

erlingfn:
The board wishes to resurrect the weekly email, to automate augment weekly. This PR creates a celery task which sends out an email containing the events with pools opening next week, recently published joblistings as well as the latest weekly if it has been written the day the email is sent. The celery task is run every sunday at 19:00 pm. 

For the first week, it will just be sent to PR and Webkom, but will in the future be sent to all students.

Fixes ABA-17


lego/apps/email/notifications.py
```diff
@@ -0,0 +1,120 @@
...
+    "lunch_presentation": "Lunchpresentasjon",
+    "alternative_presentation": "Alternativ presentasjon",
+    "course": "Kurs",
+    "breakfast_talk": "Frokostforedrag",
+    "kid_event": "KID-arrangement",
+    "party": "Fest",
+    "social": "Sosialt",
+    "other": "Annet",
+    "event": "Arrangement",
+}
```
ivarnakken, line 31: I would prefer if you'd use the mappings in `lego.apps.companies.constants.py`. We don't have a direct mapping, so you'll have to do some trickery with what we've got though. Or, move this to the file, if it turns out very ugly. The same kinda goes for `JOB_TYPE_MAPPINGS` above, please move it.

lego/apps/email/notifications.py
```diff
@@ -0,0 +1,120 @@
...
+                            "%d/%m kl. %H:%M"
+                        ),
+                    }
+                )
+
+            events.append(
+                {
+                    "title": event.title,
+                    "id": event.id,
+                    "pools": pools,
```
ivarnakken, line 99: Why not just set this to `event.pools.all()` in the first place, and remove the whole iteration thingy above? Are you not able to format the date nicely in the html directly?

lego/apps/email/tasks.py
```diff
@@ -0,0 +1,23 @@
+from structlog import get_logger
+
+from lego import celery_app
+from lego.apps.email.notifications import WeeklyNotification
+from lego.apps.users.models import AbakusGroup
+from lego.utils.tasks import AbakusTask
+
+log = get_logger()
```
ivarnakken, line 8: This seems to not be used, but I now see that we have these things laying around everywhere in our codebase, so I'll let it slide. You never know when you'll need it, I suppose?

erlingfn:
I will of course squash my commits before merging!

lego/apps/email/tasks.py
erlingfn: That is a good question. I just copied it from another task :shrug: . I could look into it if you want

lego/apps/email/notifications.py
erlingfn: Yes, I will move it. Translated events is used more for company interest than actual events, so I don't think it can be used here. I think we will benefit from having a separate dictionary or having one that is merged with this

lego/apps/email/notifications.py
erlingfn: Because django does not accept querysets as context for some reason. So you have to iterate over it and store it in an array anyways

ollfkaih (approved):
This is a good feature

lego/apps/email/notifications.py
```diff
@@ -0,0 +1,101 @@
...
+        yesterday_timestamp = timezone.now() - timedelta(days=1)
+        last_sunday_timestamp = timezone.now() - timedelta(days=7)
+
+        weekly_tag = Tag.objects.filter(tag="weekly").first()
+        # Check if weekly tag exists so it does not crash if some idiot deletes the weekly tag
+        todays_weekly = (
+            weekly_tag.article_set.filter(created_at__gt=yesterday_timestamp).first()
+            if weekly_tag
+            else None
+        )
```
LudvigHz, line 29: Hmmm, I feel like this code is kinda relying on a lot of things going right, which could go wrong. F.ex. if someone creates the weekly article before `yesterday_timestamp` (also if this task doesn't run exactly at the right day).

Would prefer that this maybe include some more leeway

lego/apps/email/notifications.py
```diff
@@ -0,0 +1,101 @@
...
+        week_number = timezone.now().isocalendar().week
+
+        events_next_week = Event.objects.filter(
+            pools__activation_date__gt=timezone.now(),
+            pools__activation_date__lt=timezone.now() + timedelta(days=7),
+        ).distinct()
+
+        filtered_events = filter(
+            lambda event: self.user.has_perm(VIEW, obj=event), events_next_week
+        )
```
LudvigHz, line 40: You _could_ also use the `filter_queryset` method on the permission handler, but this shouldn't be too big (and runs async) so it's fine I guess.

lego/apps/email/notifications.py
```diff
@@ -0,0 +1,101 @@
...
+
+        filtered_events = filter(
+            lambda event: event.get_possible_pools(self.user, True)
+            or event.is_admitted(self.user),
+            filtered_events,
+        )
+
+        joblistings_last_week = Joblisting.objects.filter(
+            created_at__gt=last_sunday_timestamp, visible_from__lt=timezone.now()
```
LudvigHz, line 49: Appreciate these variable names :1st_place_medal:

lego/apps/email/notifications.py
```diff
@@ -0,0 +1,101 @@
...
+                plain_template="email/email/weekly_mail.txt",
+                context={
+                    "events": events,
+                    "todays_weekly": ""
+                    if todays_weekly is None
+                    else todays_weekly.get_absolute_url(),
+                    "joblistings": joblistings,
+                },
+            )
+        return
```
LudvigHz, line 101: Why the empty return?

lego/apps/email/tasks.py
```diff
@@ -0,0 +1,23 @@
...
+    all_users = set(
+        AbakusGroup.objects.get(name="Webkom").restricted_lookup()[0]
+        + AbakusGroup.objects.get(name="PR").restricted_lookup()[0]
+    )
+
+    self.setup_logger(logger_context)
+
+    for user in all_users:
+        notification = WeeklyNotification(user)
+        notification.notify()
```
LudvigHz, line 23: This is, sadly, not the way to do this :/

For prod, we cant really do this at all. Considering you are using the user for permission context I understand the reasoning, but we _cannot_ send an individual email to ALL users.

You will need to make the email tailored for the "average" user and just make sure that the permissions are based on the group you are sending too (e.g. `Abakus`), and batch send the email to all users in that group. For django, this is done using https://docs.djangoproject.com/en/4.0/topics/email/#send-mass-mail.

LudvigHz: It might be fine for the testing though, seen as it is limited to the groups above for now, but I think it's a good idea to test with the "production ready" code when we're first testing.

lego/apps/users/fixtures/test_abakus_groups.yaml
```diff
@@ -545,3 +545,22 @@
...
+    logo: null
+    type: komite
+    text: ''
+    permissions: '["/sudo/admin/companyinterest/", "/sudo/admin/surveys/", "/sudo/admin/submissions/"]'
+    show_badge: true
+    active: true
+    lft: 5
+    rght: 6
+    tree_id: 18
+    level: 2
```
LudvigHz, line 566: It's not entirely obvious (we should add a header to this file), but this file is auto-generated from `test_abakus_groups.py`. So this will be overwritten the next time someone runs the generate command.

LudvigHz (changes requested):
Overall good :100: 

But there are a couple of things that needs to be looked at before we can merge this :smiling_face_with_tear:.

lego/apps/email/notifications.py
erlingfn: As far as I know, it is always written on a sunday, but I have no problem with more leeway. Would three days be sufficient?

lego/apps/users/fixtures/test_abakus_groups.yaml
erlingfn: Okay, then I will just add PR to test_abakus_groups.py

lego/apps/email/tasks.py
erlingfn: I think you have misunderstood https://docs.djangoproject.com/en/4.0/topics/email/#send-mass-mail. It is not made for sending the same email to a mass of users, but sending out _different_ emails at the same time. I agree that what I have done is not working, but we do not **not** have to send the same email to all users. Creating an email per user, then using a single call of `send_mass_email` should suffice.

LudvigHz: Yep, I did not pay attention there. You are of course correct in that we don't need to send the same email :)

ivarnakken (approved):
Looking good to me now! 💯 Excited to see this in action!

PR #3012, opened 2022-10-19
Attatch users' photo consents on event & make sure consent for event exists before registration

LudvigHz:
Previously, we checked photo_consent status for an event in the frontend
by comparing the users registered photo consents with the event
start_time.

This among other things, led to a bug where the user does not have any
photo_consents for an event if the event is in the next semester, and
the user cannot attend the event if `use_consent=True`. This fixes that
by attaching and creating photo consents for that specific event.

This PR adds `photo_consents` on the `UserEventSerializer`, and will create the relevant photo
consent objects if they don't exist. Note that we don't create the photo consents unless the user
can register for the event _and_ the event enforces photo consent usage. This is just to avoid
creating photo consents for events far in the future, as well as events that have already happened.

### DIFF
--- a/lego/apps/events/serializers/events.py
+++ b/lego/apps/events/serializers/events.py
@@ -1,5 +1,6 @@
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import transaction
+from django.http import HttpRequest
 from rest_framework import serializers
 from rest_framework.fields import CharField
 
@@ -27,7 +28,8 @@ from lego.apps.files.fields import ImageField
 from lego.apps.tags.serializers import TagSerializerMixin
 from lego.apps.users.constants import GROUP_GRADE
 from lego.apps.users.fields import AbakusGroupField
-from lego.apps.users.models import AbakusGroup
+from lego.apps.users.models import AbakusGroup, PhotoConsent
+from lego.apps.users.serializers.photo_consents import PhotoConsentSerializer
 from lego.apps.users.serializers.users import PublicUserSerializer
 from lego.utils.serializers import (
     BasisModelSerializer,
@@ -205,6 +207,7 @@ class EventReadUserDetailedSerializer(EventReadDetailedSerializer):
     spots_left = SpotsLeftField()
     pending_registration = serializers.SerializerMethodField()
     price = serializers.SerializerMethodField()
+    photo_consents = serializers.SerializerMethodField()
 
     class Meta(EventReadDetailedSerializer.Meta):
         fields = EventReadDetailedSerializer.Meta.fields + (  # type: ignore
@@ -213,15 +216,16 @@ class EventReadUserDetailedSerializer(EventReadDetailedSerializer):
             "is_admitted",
             "spots_left",
             "pending_registration",
+            "photo_consents",
         )
 
     def get_price(self, obj):
-        request = self.context.get("request", None)
+        request: HttpRequest = self.context.get("request", None)
         if request:
             return obj.get_price(user=request.user)
 
     def get_pending_registration(self, obj):
-        request = self.context.get("request", None)
+        request: HttpRequest = self.context.get("request", None)
         if not request or not request.user.is_authenticated:
             return None
 
@@ -234,6 +238,19 @@ class EventReadUserDetailedSerializer(EventReadDetailedSerializer):
             return RegistrationReadDetailedSerializer(reg).data
         except ObjectDoesNotExist:
             return None
+
+    def get_photo_consents(self, obj: Event):
+        request: HttpRequest = self.context.get("request", None)
+        if not request or not request.user.is_authenticated:
+            return []
+
+        # Only return consents for events that use consent
+        # and the user is allowed to register
+        if not obj.use_consent or len(obj.get_possible_pools(request.user)) == 0:
+            return []
+
+        pc = PhotoConsent.get_consents(request.user, time=obj.start_time)
+        return PhotoConsentSerializer(instance=pc, many=True).data
 
 
 class EventReadAuthUserDetailedSerializer(EventReadUserDetailedSerializer):
--- a/lego/apps/events/tests/test_events_api.py
+++ b/lego/apps/events/tests/test_events_api.py
@@ -19,8 +19,8 @@ from lego.apps.events.tasks import (
 )
 from lego.apps.events.tests.utils import get_dummy_users, make_penalty_expire
 from lego.apps.surveys.models import Submission, Survey
-from lego.apps.users.constants import GROUP_GRADE
-from lego.apps.users.models import AbakusGroup, Penalty, User
+from lego.apps.users.constants import GROUP_GRADE, PHOTO_CONSENT_DOMAINS
+from lego.apps.users.models import AbakusGroup, Penalty, PhotoConsent, User
 from lego.utils.test_utils import BaseAPITestCase, BaseAPITransactionTestCase
 
 _test_event_data = [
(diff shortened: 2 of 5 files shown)


eikhr (approved):
LGTM 📸

lego/apps/events/serializers/events.py
```diff
@@ -235,6 +238,17 @@ def get_pending_registration(self, obj):
         except ObjectDoesNotExist:
             return None
 
+    def get_photo_consents(self, obj):
+        request = self.context.get("request", None)
+        if not request or not request.user.is_authenticated:
+            return []
+
+        if not obj.use_consent or len(obj.get_possible_pools(request.user)) == 0:
+            return []
```
AdrianAndersen, line 247: Why is this not a single if statement? IMO looks better to have everything in one.

lego/apps/users/models.py
```diff
@@ -552,24 +552,33 @@ class PhotoConsent(BasisModel):
...
-
-        PhotoConsent.objects.get_or_create(
-            user=user,
-            year=current_year,
-            semester=current_semester,
-            domain=constants.SOCIAL_MEDIA_DOMAIN,
-        )
+        consent_time = now
+        if time is not None:
+            consent_time = time
```
AdrianAndersen, line 559:
```suggestion
        consent_time = time if time is not None else now
```

AdrianAndersen (approved):
LGTM 👍🏻 Just a bit of code style.

lego/apps/events/serializers/events.py
LudvigHz: I split it up both to make it cleaner and to separate logic.

So the first one checks if the user is defined and the second checks if the user should have consents for the event.

I can merge them but I just find short conditional statements alot easier to read and understand.

ivarnakken: I think neither solution is very elegant. I guess a comment of what you're validating here (like you're stating in the PR) would help.

ivarnakken (approved):
Makes sense :shipit: 

As always, please add some tests ❤️ I would also appreciate if you statically typed more of your code.

lego/apps/events/serializers/events.py
LudvigHz: I thought about adding a comment here, but it's very well explained in the commit message.

ivarnakken: I know, but I'd prefer to not have to git blame something to understand ...

lego/apps/users/models.py
```diff
@@ -552,24 +552,33 @@ class PhotoConsent(BasisModel):
     class Meta:
         unique_together = ("semester", "year", "domain", "user")
 
-    def get_consents(self, user):
+    def get_consents(self, user, *, time=None):
```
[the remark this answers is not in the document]
LudvigHz: The `*` param just means that all the parameters following are named, so its not possible to pass an unnamed argument for a named parameter. It just makes the code more safe, especially when refactoring.

I don't see why the name here matters really? Since its quite explanatory and is not necessarily connected to the serializer, but if you really want to I can change it.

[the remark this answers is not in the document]
ivarnakken: Cool, that makes sense. And the name is quite explanatory, indeed, but I would just like it to be perfect ..

lego/apps/users/models.py
```diff
@@ -552,24 +552,33 @@ class PhotoConsent(BasisModel):
...
-        )
+        consent_time = now
+        if time is not None:
+            consent_time = time
+        consent_semester = PhotoConsent.get_semester(consent_time)
+        consent_year = consent_time.year
+
+        # Don't create PhotoConsent objects for the past
+        if consent_time >= now:
+            PhotoConsent.objects.get_or_create(
```
erlingfn, line 565: Could it be an idea to loop over the domains in the constants in case we add more domains in the future?

LudvigHz: Yes!
I thought about it, but wanted this PR to just cover the bug it's fixing.
I could fix up all of the consent code as well, but it will make the PR larger as well as make more time.


Premature optimization is a thing, and there is no point in refactoring code before it's necessary. I think we should leave this until we _actually_ add another domain, then we can refactor and fix this code. Otherwise you just get caught in an endless refactoring and code change loop.

To quite Martin Fowler from the book _refactoring_[^1]:
> The first time you do something, you just do it. The second time you do something similar, you wince at the duplication, but you do the duplicate thing anyway. The third time you do something similar, you refactor.

[^1]: https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

LudvigHz:
> As always, please add some tests :heart: I would also appreciate if you statically typed more of your code.

I did want to, but I just winced at the thought of writing tests for this, especially considering how horrible our test framework is atm :grimacing:. (I can see how much work it is to add the tests though)

static typing I can do though, thanks for the reminder!

lego/apps/events/serializers/events.py
LudvigHz: I'll add the comment.

But I will say that avoiding `git blame` is a really bad idea. It's super useful to get the full context of a change, with the authors intention and comment on the entire change and not just a single block of code.

LudvigHz:
@ivarnakken I have added some tests for the model, as well as the event API to cover the code paths introduced here. They should cover most of the cases here. And added some types, although there wasn't really that much to type (we really need stronger type checking to be able to do this properly :/)

lego/apps/users/models.py
erlingfn: Okay, fair enough

lego/apps/events/serializers/events.py
ivarnakken: Yes, I think combining code as docs and a good commit message is the optimal solution :)

lego/apps/users/tests/test_models.py
```diff
@@ -426,5 +430,22 @@ def test_get_consents_with_prior_consents(self):
...
         self.assertEqual(len(user_consents), 2)
+
+    def test_get_consents_for_past(self):
+        """
+        It should not be possible to generate new consents for the past.
+        """
+        user_consents = PhotoConsent.get_consents(self.test_user)
+        self.assertEqual(len(user_consents), 2)
+
+        today = timezone.now()
```
ivarnakken, line 443: Kind of unnecessary variable, especially considering it's only used to calculate a different value.

lego/apps/users/tests/test_models.py
```diff
@@ -426,5 +430,22 @@ def test_get_consents_with_prior_consents(self):
...
+
+        consents = PhotoConsent.get_consents(
+            self.test_user, time=today - timedelta(weeks=52)
+        )
+        self.assertEqual(len(consents), 0)
```
ivarnakken, line 448: `assertEmpty` is much needed here, but it seems that it's not available through our current use of the unittest library?

lego/apps/events/tests/test_events_api.py
```diff
@@ -2034,3 +2034,104 @@ def test_filter_out_old(self):
...
+        event_response = self.client.get(_get_detail_url(event.id))
+
+        user_consents = PhotoConsent.get_consents(self.abakus_user)
+        self.assertEqual(
+            len(user_consents),
+            len(user_consents_before),
+            "User should have the same amount of consent objects",
+        )
+
+        self.assertEquals(
```
ivarnakken, line 2133: See comment below.

ivarnakken (approved):
Thanks for the thorough tests! 😍 Just two very small comments, which I'll leave up to you to decide what to do with 🤗

lego/apps/users/tests/test_models.py
LudvigHz: Yes, that would be nice but doesn't exist so :shrug:

PR #3020, opened 2022-10-21
Remove incorrect comma -> tuple

LudvigHz:
Commit 068e824be forgot to remove a comma, leading to a block to become
a tuple, breaking this iteration.


LudvigHz:
It might be worth looking at 068e824be to see if this happened in other places as well.

eikhr (approved):
Good catch, hard to spot

ivarnakken (commented):
Oooops 😬

PR #3033, opened 2022-10-26
Fix hard coded groups

erlingfn:
Comittees and interest groups had hard coded behavior in a lot of places, which makes it hard to add new public groups. This will enable all users to view the membership of public groups, as intended.


ivarnakken (approved):
:shipit:

LudvigHz (approved):
LGTM :shipit:

PR #3046, opened 2022-10-31
Fix return of consent for admitted users & users on waiting list

LudvigHz:
- Return consents when user is admitted to event
- Only return relevant consents for event administrate

### DIFF
--- a/lego/apps/events/models.py
+++ b/lego/apps/events/models.py
@@ -605,6 +605,11 @@ class Event(Content, BasisModel, ObjectPermissionsModel):
             return self.user_reg[0].pool is not None
         return self.registrations.filter(user=user).exclude(pool=None).exists()
 
+    def is_on_waiting_list(self, user):
+        return self.registrations.filter(
+            user=user, pool=None, status=constants.SUCCESS_REGISTER
+        ).exists()
+
     def get_registration(self, user):
         return self.registrations.filter(user=user).exclude(pool=None).first()
 
--- a/lego/apps/events/serializers/events.py
+++ b/lego/apps/events/serializers/events.py
@@ -245,10 +245,15 @@ class EventReadUserDetailedSerializer(EventReadDetailedSerializer):
             return []
 
         # Only return consents for events that use consent
-        # and the user is allowed to register
+        if not obj.use_consent:
+            return []
+
+        # If the user is not allowed to register and is not registered,
+        # there is no need for consents
         if (
-            not obj.use_consent
-            or len(obj.get_possible_pools(request.user, future=True)) == 0
+            not obj.is_admitted(request.user)
+            and not obj.is_on_waiting_list(request.user)
+            and len(obj.get_possible_pools(request.user, future=True)) == 0
         ):
             return []
 
--- a/lego/apps/events/tests/test_events_api.py
+++ b/lego/apps/events/tests/test_events_api.py
@@ -2140,3 +2140,72 @@ class EventPhotoConsentTestCase(BaseAPITestCase):
             0,
             "photoConsents should not exist on the event response",
         )
+
+    def test_event_consent_admitted(self):
+        """
+        PhotoConsents should return consents if the user is admitted
+        """
+        event = Event.objects.get(title="POOLS_NO_REGISTRATIONS")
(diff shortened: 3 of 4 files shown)


lego/apps/events/models.py
```diff
@@ -605,6 +605,11 @@ def is_admitted(self, user):
             return self.user_reg[0].pool is not None
         return self.registrations.filter(user=user).exclude(pool=None).exists()
 
+    def is_on_waiting_list(self, user):
+        return self.registrations.filter(
+            user=user, pool=None, status=constants.SUCCESS_REGISTER
+        ).exists()
```
ivarnakken, line 611: wait, how does this check if someone's on waiting list? Is the waiting list pool `None`?

lego/apps/events/serializers/events.py
```diff
@@ -247,8 +247,12 @@ def get_photo_consents(self, obj: Event):
...
         # and the user is allowed to register
         if (
-            not obj.use_consent
-            or len(obj.get_possible_pools(request.user, future=True)) == 0
+            not obj.is_admitted(request.user)
+            and not obj.is_on_waiting_list(request.user)
+            and (
+                not obj.use_consent
+                or len(obj.get_possible_pools(request.user, future=True)) == 0
+            )
```
ivarnakken, line 255: Just when I though this couldn't get any uglier ... It's impossible to read (in combination with the check above) and I'll just have to trust that it works ... Not to mention that further developing this is a pain in the ass

lego/apps/events/tests/test_events_api.py
```diff
@@ -2140,3 +2140,72 @@ def test_event_without_consent_in_future(self):
...
+        self.assertEqual(
+            len(user_consents),
+            len(PHOTO_CONSENT_DOMAINS) * 2,
+            "User should have an extra semester worths of consents",
+        )
+
+        for consent in event_response.json()["photoConsents"]:
+            self.assertIn(
+                (consent["year"], consent["semester"], consent["domain"]),
+                [(c.year, c.semester, c.domain) for c in user_consents],
```
Arashfa0301, line 2209: Maybe consent instead of c ??

ivarnakken (approved):
Tests look solid - nice 💯

lego/apps/events/tests/test_events_api.py
Arashfa0301: or user_consent...

lego/apps/events/serializers/events.py
LudvigHz: Hmm, yeah maybe you're right :grimacing:. What about this:
```python
if (
    not obj.use_consent or (
        not obj.is_admitted(request.user)
        and not obj.is_on_waiting_list(request.user)
        and len(obj.get_possible_pools(request.user, future=True)) == 0
    )
 )
```

I also wanted to extract the `not obj.is_admitted` into a separate if, but the peeps at the office preferred this.

Also, there is a bug here as you can see :upside_down_face:

Arashfa0301:
Nice job. The only note worthy thing is that there are some duplicate code in your tests. Maybe there could be something done about them?

Arashfa0301 (approved):

lego/apps/events/models.py
LudvigHz: yes. If a registration is on the waiting list, it's pool is `None`. Of course it has to be a `success` as well :)

juniwbjerde (approved):
Looks good after fixing the one if as @ivarnakken commented:)

lego/apps/events/serializers/events.py
LudvigHz: @ivarnakken I changed it to the suggestion here now. Let me know if you still think it should be different, otherwise I'll merge this ASAP.

ivarnakken: Sure, this looks better. I think `not obj.is_admitted(request.user)` should be separated though

erlingfn (approved):
Looks good!

lego/apps/events/serializers/events.py
LudvigHz: Done

PR #3080, opened 2022-11-21
Lock pools as well as event when bumping

LudvigHz:
Previously, only the event itself was locked. When a registration is registered to an event, we only lock the pool, since we don't actually care about the event unless it's merged. This means that we can see race conditions when we lock the _event_ when bumping (e.g. in `unregister()`), and a single _pool_ in `register()` while actually modifying the reg count of that same _pool_ in `unregister()`.

By locking the pools in all of these places, as well as the event while bumping, we make sure that only one of these happens at a time.

OOPS: 
I plan to try making a test to try to repro this. But not today...
Awesome if someone wants to look over and see if this makes sense in the mean time though :)

### DIFF
--- a/lego/apps/events/fixtures/test_events.yaml
+++ b/lego/apps/events/fixtures/test_events.yaml
@@ -264,7 +264,7 @@
     name: Abakusmember
     capacity: 2
     event: 10
-    counter: 2
+    counter: 1
     activation_date: "2019-09-01T13:20:30+03:00"
     permission_groups:
       - - Abakus
--- a/lego/apps/events/tasks.py
+++ b/lego/apps/events/tasks.py
@@ -2,6 +2,7 @@ import uuid
 from datetime import timedelta
 
 from django.db import IntegrityError, transaction
+from django.db.models import Q
 from django.utils import timezone
 
 import stripe
@@ -456,7 +457,8 @@ def check_events_for_registrations_with_expired_penalties(self, logger_context=N
         with transaction.atomic():
             locked_event = Event.objects.select_for_update().get(pk=event_id)
             if locked_event.waiting_registrations.exists():
-                for pool in locked_event.pools.all():
+                locked_pools = locked_event.pools.select_for_update().all()
+                for pool in locked_pools:
                     if pool.is_activated and not pool.is_full:
                         for _ in range(locked_event.waiting_registrations.count()):
                             locked_event.check_for_bump_or_rebalance(pool)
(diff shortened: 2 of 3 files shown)


lego/apps/events/tasks.py
```diff
@@ -475,7 +477,11 @@ def bump_waiting_users_to_new_pool(self, logger_context=None):
     )
     for event_id in events_ids:
         with transaction.atomic():
-            locked_event = Event.objects.select_for_update().get(pk=event_id)
+            locked_event = (
+                Event.objects.select_related("pools")
```
ivarnakken, line 481: Why is the `select_related("pools")` needed here? My understanding here is that the `of` arg will lock related objects, in this case `pools`, so wouldn't that be sufficient?

LudvigHz: Oops, yeah I initially used a `select_related` to lock the pools but realized that that won't work, and we have to use the `of` parameter here.

LudvigHz: I had to change this as well, since we apparently can't lock rows of related objects like this. We can only lock related objects included in the query, which pools are not here. So I instead opted to just lock the event, and then lock all the pools. I _think_ this should be fine, since the order of locking shouldn't matter since we only check the reg counter when the event is merged, and the lock should be acquired in all those places.

lego/apps/events/fixtures/test_events.yaml
```diff
@@ -264,7 +264,7 @@
     name: Abakusmember
     capacity: 2
     event: 10
-    counter: 2
+    counter: 1
```
LudvigHz, line 267: This was incorrect, since the test making sure that the task for ensuring the counter is correct of course did not work. Now that the task is fixed, the tests now failed on this event of course.

This was prob just copy-pasted, so should not have any negative effects.

lego/apps/events/tasks.py
```diff
@@ -477,7 +479,8 @@ def bump_waiting_users_to_new_pool(self, logger_context=None):
         with transaction.atomic():
             locked_event = Event.objects.select_for_update().get(pk=event_id)
             if locked_event.waiting_registrations.exists():
-                for pool in locked_event.pools.all():
+                locked_pools = locked_event.pools.select_for_update().all()
+                for pool in locked_pools:
```
LudvigHz, line 483: The locking order here should be fine, since we check for `pool.is_full` below. Had we had `event.is_full` (when merged) this would cause race conditions as well.

We should be wary to make sure we have this order as well everywhere, otherwise we could end up with deadlocks :grimacing:

lego/apps/events/tasks.py
```diff
@@ -577,13 +580,15 @@ def check_that_pool_counters_match_registration_number(self, logger_context=None
     self.setup_logger(logger_context)
 
     events_ids = Event.objects.filter(
-        start_time__gte=timezone.now(), merge_time__gte=timezone.now()
+        Q(start_time__gte=timezone.now()),
```
ivarnakken, line 583: Is this necessary?

LudvigHz: Well, otherwise we'd be checking all old events? This makes sure we only test for events that are in the future.

ivarnakken: oh shoot I'm sorry, I read `merge_time__gte` and thought it was a duplicate of the one below ..

PR #3081, opened 2022-11-22
Make not-accepted meetings transparent in calendar

ollfkaih:
Fixes an annoyance of mine, kinda™️    
bilde
Appears white (test123) and not dashed in google calendar, but this seems to be the proper way to do this

### DIFF
--- a/lego/apps/ical/utils.py
+++ b/lego/apps/ical/utils.py
@@ -16,9 +16,9 @@ def add_events_to_ical_feed(
         add_event_to_ical_feed(feed, event, price, title, ical_starttime, ical_endtime)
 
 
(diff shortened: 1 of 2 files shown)


ivarnakken (approved):
Good shit 💯

norbye (approved):

LudvigHz:
I just had a quick lock at the iCalendar protocol, and would it not be a lot better to use the `STATUS` field[^1]? This will also make it work better for more calendar clients. I'm guessing the `transparency` field here is a gCal specific thing?

[^1]: https://icalendar.org/iCalendar-RFC-5545/3-2-12-participation-status.html

PR #3087, opened 2022-11-27
Enforce Python 3.11 🚄 

ivarnakken:
Okay, let's be agile here and give Python 3.11 a chance. Crossing fingers. 

`lego-testbase` is upgraded [here](https://github.com/webkom/lego-testbase/pull/6).


LudvigHz (approved):
Lets go!!!
:shipit: :bullettrain_side:

PR #3088, opened 2022-11-27
Fix weekly mail when email is empty

erlingfn:
The behavior for not sending an email when there is nothing to put in the weekly mail was broken. Unfortunately the tast was also broken, so this error was not discovered. This fixes both the behavior and the test.


eikhr (approved):
lgtm :)

LudvigHz (approved):

PR #3089, opened 2022-11-28
Bump mypy to version 0.991

ivarnakken:
Also, in accordance, bump `django-stubs` to v1.13 and `djangorestframework-stubs` to v1.8.

I want to get this in before I start typing a bunch of stuff.

### DIFF
--- a/lego/apps/events/models.py
+++ b/lego/apps/events/models.py
@@ -1,5 +1,6 @@
+from __future__ import annotations
+
 from datetime import timedelta
-from typing import Union
 
 from django.conf import settings
 from django.core.exceptions import ValidationError
@@ -376,7 +377,7 @@ class Event(Content, BasisModel, ObjectPermissionsModel):
                 if follow_event_item:
                     follow_event_item.delete()
 
-    def check_for_bump_or_rebalance(self, open_pool: "Pool"):
+    def check_for_bump_or_rebalance(self, open_pool: Pool):
         """
         Checks if there is an available spot in the event.
         If so, and the event is merged, bumps the first person in the waiting list.
@@ -399,7 +400,7 @@ class Event(Content, BasisModel, ObjectPermissionsModel):
                         return self.bump(to_pool=open_pool)
                 self.try_to_rebalance(open_pool=open_pool)
 
-    def bump(self, to_pool: "Pool" = None):
+    def bump(self, to_pool: Pool | None = None):
         """
         Pops the appropriate registration from the waiting list,
         and moves the registration from the waiting list to `to pool`.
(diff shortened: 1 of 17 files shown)


lego/apps/users/models.py
```diff
@@ -1,3 +1,4 @@
+# mypy: ignore-errors
```
ivarnakken, line 1: It's impossible to type to due a weird mypy bug, and I'm not gonna spend another month on this ...

setup.cfg
```diff
@@ -16,9 +16,14 @@ combine_as_imports = true
...
 [mypy]
 python_version = 3.11
-plugins = mypy_django_plugin.main,mypy_drf_plugin.main
-ignore_missing_imports = True
-allow_untyped_globals = True
+plugins = mypy_django_plugin.main, mypy_drf_plugin.main
+ignore_missing_imports = true
+allow_untyped_globals = true
+# TODO: Enable this, but it requires a lot of work (~600 errors)
+check_untyped_defs = false
```
ivarnakken, line 23: Didn't know this was a thing, but we have been suppressing roughly 600 errors because of it. Should perhaps enable and set an allowed error amount (like we've done for typescript in the webapp)? Thoughts?

lego/apps/events/tasks.py
```diff
@@ -6,7 +6,7 @@
 from django.utils import timezone
 
 import stripe
-from celery import chain
+from celery.canvas import chain
```
LudvigHz, line 9: Why has this import changed?

setup.cfg
ivarnakken: Also, we should configure `mypy` in general a lot better here! There's many features which are currently not utilizing.

lego/apps/users/models.py
LudvigHz: I'm guessing it's the `UserManager` that is causing problems? I spent some time on it as well, and it's a goddam nightmare, although I _believe_ it should be possible to fix. 

Do you have a reference to an issue report or similar? Would be nice to have that so we can check on it once in a while if this is fixed/simplified.

LudvigHz (approved):
Nice :100: 
Looks good. (actually way fewer fixes than I would have thought, disregarding the manager hell).

setup.cfg
LudvigHz: I agree. Setting the limit to what it is now at least enforces all new code to be typed :)
When we set up mypy last time, it was just the bootstrapping, and we ignored all types just to make it pass CI. But I think enabling some of the actual nice features would be nice.

ivarnakken:
> I'm guessing it's the UserManager that is causing problems?

Yes.

> Do you have a reference to an issue report or similar? Would be nice to have that so we can check on it once in a while if this is fixed/simplified.

I do not currently have one, but I vaguely remember some related issues when I worked on this during summer.

ivarnakken:
> (actually way fewer fixes than I would have thought, disregarding the manager hell).

That's because most are suppressed (see comment above regarding `check_untyped_defs`).

lego/apps/events/models.py
```diff
@@ -1,5 +1,6 @@
+from __future__ import annotations
```
ivarnakken, line 1: I thought this would finally be mandatory in Python 3.11, but that was not the case ..

PR #3126, opened 2022-12-17
Type improvements

ivarnakken:
- Configure `mypy` more thoroughly, and allow it to fail on CI
- Statically type event models (~140 errors down)

---

Resolves ABA-182


lego/apps/events/models.py
```diff
@@ -755,40 +793,40 @@ class Pool(BasisModel):
     class Meta:
         ordering = ["id"]
 
-    def delete(self, *args, **kwargs):
+    def delete(self, *args: Any, **kwargs: Any) -> tuple[int, dict[str, int]]:
         if not self.registrations.exists():
-            super().delete(*args, **kwargs)
+            return super().delete(*args, **kwargs)
```
ivarnakken, line 798: Assuming this won't affect any logic? `super().delete()` returns a tuple further up, making it wrong to say this returns `None`, thus making mypy complain that there's no return statement.

lego/apps/events/models.py
```diff
@@ -871,20 +909,20 @@ def should_notify(self, time=None):
...
         return False
 
-    def set_payment_success(self):
+    def set_payment_success(self) -> Registration:
         self.payment_status = constants.PAYMENT_MANUAL
         self.save()
         return self
 
-    def set_presence(self, presence):
+    def set_presence(self, presence: str) -> None:
```
ivarnakken, line 917: `presence` must be in `constants.PRESENCE_CHOICES`, but I couldn't get to type it any better than this. Would love some input.

lego/apps/users/models.py
```diff
@@ -468,7 +470,7 @@ def restricted_lookup(self):
     def announcement_lookup(self):
         return [self]
 
-    def unanswered_surveys(self):
+    def unanswered_surveys(self) -> list:
```
ivarnakken, line 473: Actually returning `list[Survey]`, but the `Survey` model is imported inside of the function, so this'll have to do.

setup.cfg
```diff
@@ -19,8 +19,25 @@ python_version = 3.11
 plugins = mypy_django_plugin.main, mypy_drf_plugin.main
 ignore_missing_imports = true
 allow_untyped_globals = true
-# TODO: Enable this, but it requires a lot of work (~600 errors)
```
ivarnakken, line 22: It's actually over 5000 errors btw.

lego/apps/events/models.py
```diff
@@ -532,7 +565,9 @@ def add_to_waiting_list(self, user):
             },
         )[0]
 
-    def pop_from_waiting_list(self, to_pool: Pool | None = None) -> Registration | None:
+    def pop_from_waiting_list(
+        self, to_pool: Optional[Pool] = None
+    ) -> Registration | None:
```
ivarnakken, line 570: Not sure if I like `model | None` or `Optional[model]` best. All functions returning something optional uses the pipe, while variables and args use `Optional`. Let me know what you think.

lego/apps/users/models.py
LudvigHz: You can forward reference types by making the type hint a string: `"List[Survey]"`

See https://peps.python.org/pep-0484/#the-problem-of-forward-declarations

lego/apps/events/models.py
LudvigHz: I think it's better to stick to one thing. And I'm leaning towards `Pool | None` just because it's a bit more verbose

lego/utils/models.py
```diff
@@ -21,7 +23,7 @@ def __init__(self, *args, **kwargs):
                 if isinstance(v, abakus_cached_property):
                     self._cached_properties[k] = v
 
-    def save(self, *args, **kwargs):
+    def save(self, *args: Any, **kwargs: Any) -> None:
```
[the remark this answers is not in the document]
ivarnakken: There's no other way of typing this?

lego/apps/users/models.py
ivarnakken: Yea I tried `list["Survey"]`, but my interpreter was behaving weirdly, giving a warning that `"Survey" is not defined`. Will look into.

lego/utils/models.py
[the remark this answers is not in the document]
LudvigHz: Yes :)
It is a big job to fix, and I do agree that re-typing these is very tedious so if there is a way to tell mypy that the method should have the same signature as the parent class than that would solve a lot of problems.

I  couldn't find an easy way to use `*args, **kwargs` when overriding a typed method in mypy though so it might be more trouble than it's worth... Thats python typing for you :3rd_place_medal:.

For reference, here is the method in django: https://github.com/django/django/blob/04fdf719331dde6b0f4e7cdc445be4d4278f3ec4/django/db/models/base.py#L767-L769
and the type signature in django-stubs: https://github.com/typeddjango/django-stubs/blob/a7a15185a5ede9b487cf40ec02da08527b81afc5/django-stubs/db/models/base.pyi#L46-L52

lego/apps/events/models.py
LudvigHz: I would make `constants.PRESENCE_CHOICES` an `Enum` and type everything accordingly. That would give type safety and make everything alot cleaner ;)

LudvigHz (approved):

lego/apps/users/models.py
ivarnakken: `Pylance` is still telling me that `"Survey" is not defined`, but I can live with that.

lego/utils/models.py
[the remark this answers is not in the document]
ivarnakken: Yup, very tedious work, so I think I can leave that for later, or just wait for `mypy` to support inheritance 🙄 .

lego/apps/users/models.py
ivarnakken: Actually, `flake8` is now upset, so `list` it is.

PR #3140, opened 2022-12-22
Implement PDM into our project

ivarnakken:
Our requirements folder has been a mess for many years now, and it's time to introduce a package and dependency manager. I tried to compare the different options out there, and concluded that PDM was a good fit. It's much faster than its competition, and reminds me more of npm.

**Key points**
- The introduction of a lock file. This will improve efficiency and ensure that everyone is working with the right dependencies.
- PDM makes sure you're always installing into a virtual environment.
- Sadly, PDM is not supported by dependabot, but I'll try to find a different solution (read ABA-184 for more info).

---

Resolves ABA-184


pyproject.toml
```diff
@@ -0,0 +1,118 @@
...
+requires-python = ">=3.11"
+license = {text = "MIT"}
+
+[project.urls]
+Homepage = "https://lego.abakus.no"
+Repository = "https://github.com/webkom/lego"
+Documentation = "https://lego.abakus.no/docs/"
+ 
+[tool.pdm.scripts]
+start = "python manage.py runserver"
```
ivarnakken, line 68: What do people think of this? I kind of like the recemplance of npm/yarn

LudvigHz: I kinda don't like it tbh. I think it hides what's going on too much. It's a bit weird since in the django/python world, django and `manage.py` by extension is sort of the script manager, and hiding that behind pdm is unnecessary and strange.

Also, I would like to vouch for the Makefile. It's just more universal, easier to read and edit and faster. I also kind of like having it the other way round, where you have Makefiles as a wrapper around everything else that _needs_ to be handled by pdm, npm or whatever else. (this article sums it up pretty well: https://www.yieldcode.blog/post/why-you-should-adpot-makefile-in-all-of-your-projects)

pyproject.toml
```diff
@@ -0,0 +1,118 @@
...
+    # Logging/errors
+    "structlog==21.5.0",
+    "structlog-sentry==1.4.0",
+    "sentry-sdk==1.12.0",
+    "analytics-python==1.2.9",
+    # Docs
+    "Pygments==2.13.0",
+    "Markdown==3.4.1",
+    "coreapi==2.3.3",
+    "aiosmtpd==1.4.2",
```
LudvigHz, line 57: the first three here are used for docs but are not in the `docs` group. Also, `aiosmtpd` should be under base deps or perhaps an email group.

pyproject.toml
```diff
@@ -0,0 +1,118 @@
+[project]
+authors = [
+    {name = "Webkom", email = "webkom@abakus.no"},
+]
+dependencies = [
```
LudvigHz, line 5: Some of the deps here are not direct, but were included in the requirements files as a way to lock sub-deps. As we have a lockfile now, these should not be here anymore.

tox.ini
```diff
@@ -47,25 +46,26 @@ passenv =
     DATABASE
     CACHE
 commands =
-    python manage.py migrate
-    python manage.py missing_migrations
+    pdm run python manage.py migrate
+    pdm run python manage.py missing_migrations
```
LudvigHz, line 50: What's the reason for the `pdm run`  before all of these? The whole point of tox is to create an isolated environment already, but I'm not familiar with pdm and tox+pdm so please enlighten me.

README.md
```diff
@@ -12,30 +12,37 @@
...
 
 ```bash
-# Initial setup (only need to once)
 $ git clone git@github.com:webkom/lego.git && cd lego/
-$ python3 -m venv venv
 $ echo "from .development import *" > lego/settings/local.py
-$ source venv/bin/activate
-$ pip install -r requirements/dev.txt
+$ pdm install
+$ source .venv/bin/activate
```
LudvigHz, line 23: hmm, not a big deal, but I kinda like the non-hidden venv-folder

.drone.yml
```diff
@@ -10,16 +10,19 @@ trigger:
 
 steps:
   - name: setup
-    image: abakus/lego-testbase:python3.11
+    image: abakus/lego-testbase:latest
```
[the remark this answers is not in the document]
LudvigHz: Yeah, please tag the image with something and don't use latest

.drone.yml
```diff
@@ -10,16 +10,19 @@ trigger:
...
     when:
       event: push
     environment:
       LANG: C.UTF-8
+      CACHE: cache
```
LudvigHz, line 18: what does this do?

LudvigHz:
Since I (and probable everyone else) are not familiar with pdm at all, could you sum up the differences between it and poetry and pip. And why you think pdm is the better one?

.drone.yml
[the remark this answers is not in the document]
ivarnakken: Okay so first I had problems with drone refusing to pull the new image tagged with python3.11 so I had to temporarily change it, but then I kind of realized that I prefer latest. It's `lego-testbase` and supposed to be tailored to this repository, so it makes sense to use latest. Tagging with the python version makes less sense now that the image contains other packages as well (pdm and codecov).

pyproject.toml
ivarnakken: I agree that we should use `manage.py`.

I don't see why we need a Makefile when we have this integrated? make commands are _much_ slower. E.g. try to compare `pdm format` and `make fixme`. fixme requires docker images to run and installs everything without using the cache. So, faster, no. Easier to edit? They are almost never edited and how is this any more difficult?

README.md
ivarnakken: I kinda do as well, but for some reason [`pdm config venv.backend [virtualenv|venv|conda]`](https://pdm.fming.dev/latest/usage/venv/) does not work ...

But, if you already have a `venv`, it will use that instead. So you still get to use that, but just not by default.

.drone.yml
ivarnakken: Not sure yet. I was just testing because the cache was behaving weirdly.

tox.ini
ivarnakken: `pdm run` is supposed to be added to the front implicitly, but for some reason it does not ... Found this related issue https://github.com/pdm-project/tox-pdm/issues/12

ivarnakken:
>  could you sum up the differences between it and poetry

PDM and poetry are very similar. They are both better than pip-tools, uses a lock file and a `pyproject.toml` file for conf (similar to package.json for npm). But, what gives PDM an edge is that it's faster, PEP compliant and growing faster. To me, PDM just seems like a better option, but I am all ears.

README.md
ivarnakken: Updated the docs to clarify this ^

pyproject.toml
ivarnakken: Yes, but they are not needed for the `docs` step in our CI, right? Thought this would speed things up with `tox-pdm`, but I might be wrong. It's difficult to measure performance though, as the drone cache behaves weirdly. Gonna investigate more.

Regarding `aiosmtpd`, you're right, moving it now.

ivarnakken: @LudvigHz, are you sure `coreapi` is part of the `docs` group? I don't immediately see the correlation.

pyproject.toml
ivarnakken: I only find `sentry-sdk` to be a dep of `structlog-sentry` (with no required versions), `Pygments` to be a dep of `sphinx`. Do you want these to be removed? And are you aware of any more?

pyproject.toml
SmithPeder: > DEPRECATION NOTICE: Use of CoreAPI-based schemas were deprecated with the introduction of native OpenAPI-based schema generation as of Django REST Framework v3.10. See the [Version 3.10 Release Announcement](https://www.django-rest-framework.org/community/3.10-announcement/) for more details.

And since you're running `djangorestframework==3.14.0` maybe it would be an idea to just get rid of it 🤷🏻 

https://github.com/webkom/lego/blob/84fc5e728db4fe218d8481a4adf7e87f4a7ecd5b/lego/settings/rest_framework.py#L23

ivarnakken: Nice catch, @SmithPeder!

ivarnakken:
Closed in favor of #3281

PR #3155, opened 2023-01-02
Include `is_priced` in the event read serializer

ivarnakken:
Used by the webapp on /events/{id}/administrate/attendees. If you do a hard refresh here, `is_priced` will become undefined and the column of who's paid will vanish.

We have so many event serializers, and it can get confusing, so I might be missing something as to why stuff like this isn't included in the serializer from beforehand?

### DIFF
--- a/lego/apps/events/serializers/events.py
+++ b/lego/apps/events/serializers/events.py
@@ -96,6 +96,7 @@ class EventReadSerializer(
             "activation_time",
             "is_admitted",
             "survey",
+            "is_priced",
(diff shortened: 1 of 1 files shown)


ingraso:
Great! I remember noticing this bug when implementing the sorting of the column, but didn't create an issue / fix it :grimacing: But not sure why it wasn't in the serializer before :woman_shrugging: Anyways, great work :rocket:

LudvigHz (approved):

PR #3158, opened 2023-01-03
Fix photo consent tests

ivarnakken:
Tests were failing because it's the beginning of a new semester (january), and it's not enough to add 6 months to get to the next. Adding weeks instead of months is not deterministic.

This is because we distinguish between spring and autumn semester [by comparing the month to 7](https://github.com/webkom/lego/blob/master/lego/apps/users/models.py#L586). This was tested for December 31. and January 1.

---

Resolves ABA-196

### DIFF
--- a/lego/apps/events/tests/test_events_api.py
+++ b/lego/apps/events/tests/test_events_api.py
@@ -8,6 +8,7 @@ from django.utils import timezone
 from rest_framework import status
 
 import stripe
+from dateutil.relativedelta import relativedelta
 from djangorestframework_camel_case.render import camelize
 
 from lego.apps.events import constants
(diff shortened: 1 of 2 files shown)


lego/apps/events/tests/test_events_api.py
```diff
@@ -8,6 +8,7 @@
 from rest_framework import status
 
 import stripe
+from dateutil.relativedelta import relativedelta
```
LudvigHz, line 11: Why not just use `django.utils.timezone.timedelta`?

LudvigHz (approved):
LGTM :shipit:
I f**in hate time and computers...

lego/apps/events/tests/test_events_api.py
ivarnakken: `timedelta` does not support months, weeks max. Probably why you ended up doing it the hacky way.

requirements/mypy.txt
```diff
@@ -11,3 +11,4 @@ types-PyYAML==6.0.12.2
 types-requests==2.28.11.5
 types-six==1.16.21.4
 types-urllib3==1.26.25.4
+types-python-dateutil
```
ivarnakken, line 14: `mypy` forced me to add ths. I think we can agree that we don't need to specify a version on these, right ...?

PR #3198, opened 2023-02-01
Add plausible source to weekly email links

erlingfn:
Add some source query params to track if people click the links in our weekly mails. Nice to check if people actually read them, and if the lead to people checking out weekly, events or joblistings.

### DIFF
--- a/lego/apps/email/tasks.py
+++ b/lego/apps/email/tasks.py
@@ -23,6 +23,10 @@ from lego.utils.tasks import AbakusTask
 log = get_logger()
 
 
+def add_source_to_url(url):
+    return f"{url}?utm_source=WeeklyMail&utm_campaign=Email"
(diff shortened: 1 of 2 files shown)


lego/apps/email/tasks.py
```diff
@@ -23,6 +23,10 @@
 log = get_logger()
 
 
+def add_source_to_url(url):
+    return f"{url}?source=WeeklyMail"
```
LudvigHz, line 27: You _could_ also add `utm_campaign=Email` (case sensitive) so that we can see how many use email in general and not just the weeklyMail.

LudvigHz (approved):
LGTM

ivarnakken (approved):

LudvigHz (approved):

PR #3200, opened 2023-02-02
Improve weekly mail

erlingfn:
Improve the css of the joblistings table. Hard code date and times being displayed in Norwegian time.

### DIFF
--- a/lego/apps/email/tasks.py
+++ b/lego/apps/email/tasks.py
@@ -1,4 +1,5 @@
 from datetime import timedelta
+from zoneinfo import ZoneInfo
 
 from django.conf import settings
 from django.template.loader import render_to_string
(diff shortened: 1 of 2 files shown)


lego/apps/email/tasks.py
```diff
@@ -4,6 +4,7 @@
 from django.template.loader import render_to_string
 from django.utils import timezone
 
+import pytz
```
ivarnakken, line 7: Support for `pytz` was deprecated from [Django v4](https://docs.djangoproject.com/en/4.1/releases/4.0/#zoneinfo-default-timezone-implementation). Use `zoneinfo` instead.

ivarnakken (changes requested):
💯 Word break was much needed

LudvigHz (approved):
LGTM (except for the `tzinfo` ofc)

norbye (approved):

ivarnakken (approved):

PR #3214, opened 2023-02-09
Fix event-following bug by adding current user following status to event serializer

eikhr:
There is a bug where the webapp does not correctly load the following status of the current user, if an event has a large number of followers. It is caused by pagination on the event follower endpoint, but I figured this was a better solution than loading all pages in the webapp.

This adds a field "following" on `EventReadUserDetailedSerializer` that is `false` when the current user is not following the event, and the id of the relevant `EventFollow` object if the user is following the event.

https://github.com/webkom/lego-webapp/pull/3541 implements the needed front-end changes

### DIFF
--- a/lego/apps/events/fields.py
+++ b/lego/apps/events/fields.py
@@ -79,6 +79,17 @@ class IsAdmittedField(serializers.Field):
         return value.is_admitted(request.user)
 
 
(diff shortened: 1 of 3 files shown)


ivarnakken (approved):
lgtm :shipit:

PR #3215, opened 2023-02-10
Add reactions to meetings

jonasdeluna:
https://github.com/webkom/lego-webapp/pull/3548


LudvigHz (approved):
LGTM

I see now that the `get_reactions_grouped` ideally should be done in a query/SQL and not like this in python. Would speed things up a lot :)

PR #3220, opened 2023-02-13
Change author on articles from one to many

jonasdeluna:
image


lego/apps/articles/models.py
```diff
@@ -13,6 +17,7 @@ class Article(Content, BasisModel, ObjectPermissionsModel):
     youtube_url = CharField(
         max_length=200, default="", validators=[youtube_validator], blank=True
     )
+    author = ManyToManyField(User)
```
LudvigHz, line 20: This should be `authors`, no? (i.e. plural)

lego/apps/articles/serializers.py
```diff
@@ -82,7 +87,8 @@ class PublicArticleSerializer(TagSerializerMixin, BasisModelSerializer):
     cover_placeholder = ImageField(
         source="cover", required=False, options={"height": 30, "filters": ["blur(20)"]}
     )
-    author = PublicUserSerializer(read_only=True, source="created_by")
+    #    author = serializers.ListField(child=PublicUserSerializer(read_only=True))
```
LudvigHz, line 90: same here

LudvigHz (changes requested):

lego/apps/articles/migrations/0012_rename_author_article_authors.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 4.0.9 on 2023-02-28 16:34
+
```
LudvigHz, line 2: You should delete the two migrations you have here and create a new one. The reason is that we don't want unnecessary migrations in prod. You can just delete both of there, reset your db locally and run makemimgrations again to fix it :)

lego/apps/articles/serializers.py
```diff
@@ -23,6 +24,9 @@ class DetailedArticleSerializer(TagSerializerMixin, BasisModelSerializer):
     content_target = CharField(read_only=True)
     content = ContentSerializerField(source="text")
     reactions_grouped = serializers.SerializerMethodField()
+    authors = PublicUserField(
+        queryset=User.objects.all(), allow_null=True, required=False, many=True
+    )
```
LudvigHz, line 29: Do we want to allow null?

LudvigHz (changes requested):
Nice :ok_hand:. Just a few small things she this is good to go

ivarnakken (approved):
Awesome! I'm excited to get this in 🚀 

Just fix up Ludvig's tiny comment and we're good!

jonasdeluna:
Tests are failing is that supposed to happen

PeterJFB:
> Tests are failing is that supposed to happen

Yess
image


Failing likely since the tests now need to use the `authors`-field instead of `author`. All of them use the `get_data_with_author` function in `test_articles_api.py`, changing it to an array should fix it :)

PR #3227, opened 2023-02-21
Fix incorrect waiting list sorting

LudvigHz:
On some events, usually with many users in the waiting list, the waiting
list is consistently wrong for all users (i.e. the sorting is the same,
incorrect for all users) when admitted to the event.

I'm not 100% sure why this works, but changing the query for the
prefetching on the event retrieve endpoint consistently gives the
correct sorting.


norbye (approved):
Hard to give a good review as I don't really understand this either, but I'm all for deploying it and checking how it works in prod 🚀

PR #3229, opened 2023-02-26
Fix bug sending 10+ mails to all users

erlingfn:
The restricted lookup returns a list containing each user multiple times. This results in each user receiving multiple copies of each weekly mail(I think). This sends out the weekly mail to each user just once, and also just does it for students.

### DIFF
--- a/lego/apps/email/tasks.py
+++ b/lego/apps/email/tasks.py
@@ -119,8 +119,8 @@ def send_weekly_email(self, logger_context=None):
 
     week_number = timezone.now().isocalendar().week
 
-    # Set to just PR and Webkom for testing purposes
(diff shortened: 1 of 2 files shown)


lego/apps/email/tasks.py
```diff
@@ -120,7 +120,7 @@ def send_weekly_email(self, logger_context=None):
     week_number = timezone.now().isocalendar().week
 
     # Set to just PR and Webkom for testing purposes
```
ivarnakken, line 122: Might want to remove this comment as well

ivarnakken (approved):
:shipit: lgtm

LudvigHz (approved):

eikhr (approved):

PR #3230, opened 2023-02-26
Update deprecated tox config

erlingfn:
The whitelist_externals option has been deprecated and must be replaced with allowlist_externals. This will allow us to run the tests in drone again

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -18,7 +18,7 @@ commands =
 [testenv:tests]
 deps =
     -r{toxinidir}/requirements/dev.txt
-whitelist_externals =
+allowlist_externals =
     /usr/bin/make
 setenv =
     DJANGO_SETTINGS_MODULE=lego.settings


ivarnakken:
It's PDM that's causing the tests step to fail

erlingfn:
Im pretty sure this is a problem as well. Because I could not run tox locally

ivarnakken:
> Because I could not run tox locally

It works for me? And it runs fine in drone (usually, but sometimes PDM affects it lol)

ivarnakken:
#3229 passed now.

erlingfn:
It literally fails because of that line in the config 
image

ivarnakken:
> It literally fails because of that line in the config

Yes, but it's PDM that's raising it. It's a problem with PDM.

erlingfn:
I know that due to fuckups in the recent ish past, my code may not be the most trustworthy, but I think I am onto something now

ivarnakken:
> I know that due to fuckups in the recent ish past, my code may not be the most trustworthy, but I think I am onto something now

haha, yes. I'm sure this is good though. I was just pointing out that the CI step fails because if PDM. Not sure what the point of it even is 🤷🏼

LudvigHz (approved):

PR #3232, opened 2023-03-07
Fix null-value errors for exporing CompanyInterest info

Arashfa0301:
Fix null-value errors for exporing CompanyInterest info. The server couldn't find the latest model attributes because old company interest does not have the latest attributes.


ivarnakken (approved):

erlingfn (approved):
LGTM

PR #3248, opened 2023-03-17
Create new serializer for allergies

itsisak:
Removed "allergies" from the AdministrateUserSerializer and created a new serializer with allergies, so the allergies information only will be available for the owner of the event in the allergies tab. 

On hold until I finish the appropriate changes to frontend to avoid any errors.


lego/apps/users/serializers/users.py
```diff
@@ -106,19 +106,29 @@ class AdministrateUserSerializer(PublicUserSerializer):
...
     class Meta(PublicUserSerializer.Meta):
-        fields = PublicUserSerializer.Meta.fields + (  # type: ignore
-            "abakus_groups",
-            "allergies",
-        )
+        fields = PublicUserSerializer.Meta.fields + ("abakus_groups",)  # type: ignore
 
 
-class AdministrateUserExportSerializer(PublicUserSerializer):
+class AdministrateUserAllergiesExportSerializer(PublicUserSerializer):
```
ivarnakken, line 112: Why are there two serializers with `allergies` now? Does it not make more sense to keep `AdministrateUserExportSerializer`, and simply extend that to be `AdministrateUserAllergiesSerializer`?

ivarnakken (commented):
Looks good, but I'm wondering if two allergy serializers isn't the best way.

lego/apps/events/fixtures/test_events.yaml
```diff
@@ -269,6 +283,17 @@
     permission_groups:
       - - Abakus
 
+- model: events.Pool
+  pk: 9
+  fields:
+    name: Abakusmember
+    capacity: 1 
+    event: 13
+    counter: 1 #what does this do?
```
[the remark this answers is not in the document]
itsisak: Thank you! Forgot to remove the comment :/

lego/apps/users/serializers/users.py
itsisak: The point is that when getting the allergies, we dont need the information that the export serializer gives (email and phone number). So the point was to simply send as little information as needed for those different scenarios

ivarnakken: But as it's now, you can access allergies through `EventAdministrateAllergiesExportSerializer` even if you're not the creator of the event?

And then, why change the name?

itsisak: Im changing back to normal now

itsisak: Actually I have to add 'allergies' back to the export serializer because if not it is not possible to get allergies if use_contact_tracing is true. 

Is it best to let it still be called just AdministrateUserExportSerializer or is it better to change it to AdministrateUserAllergiesExportSerializer (as I did in the original commit) since it then "adds" on to the allergies serializer?

ivarnakken (approved):
Looks perfect now 🚀 Good job!

PR #3254, opened 2023-03-21
Unfollow event when unregistering

ivarnakken:
With tests! 
Also, the `unregister()` method was never actually tested ...


LudvigHz (approved):
LGTM :shipit:

PR #3255, opened 2023-03-21
Feide verification

LudvigHz:
Fixes ABA-520

### TODO

- [x] Endpoint to get user groups from feide/FS
- [x] User model updates and logic to get verification status (atm the idea is to only need to verify once, but reverifying to set correct groups would be nice to have)
- [x] Methods to validate Abakus membership from feide groups, and set correct group memberships etc


lego/apps/users/models.py
```diff
@@ -434,7 +437,13 @@ def profile_picture(self, value):
...
 
     def is_verified_student(self):
-        return self.student_username is not None
+        if self.student_username is None:
+            return False
+
+        if timezone.now() - self.student_verified_at > timezone.timedelta(
+            days=constants.STUDENT_VERIFICATION_DAYS
+        ):
+            return False
```
ivarnakken, line 446: This never returns `True`

lego/apps/users/models.py
```diff
@@ -334,6 +334,9 @@ class User(
         validators=[student_username_validator, ReservedNameValidator()],
         error_messages={"unique": "A user has already verified that student username."},
     )
+    student_verified_at = models.DateTimeField(
+        "Date last verfied student status.", null=True, default=None
+    )
```
ivarnakken, line 339: Missing a migration

ivarnakken (commented):
Looks clean and good. 💯 

What's left to do?

LudvigHz:
> What's left to do?

Most of this PR :sweat_smile:. I'll update the description with some checkboxes so it's more clear :)

LudvigHz:
_Slightly_ more left here to polish it off, but I can take reviews on this now :)

LudvigHz:
A few things left to answer on this:

### Unique FEIDE users

Right now, there is no check for the same FEIDE account on multiple users. This means that a single FEIDE user can be used to register several users. This would be quite simple to fix by saving the user ID and un-verifying users if they use the same FEIDE account on another user f.ex.


### What to do with current students?

Currently, they keep their grade, but will have `is_verified_student() = None`. This should mean that any current users can register for events etc. but might be some funky stuff with the verified property (although I don't think it's used anywhere else).

Some options are:
- Keep it as is. Current students will keep thier grade
- Use the `is_verified` property in event registration and inform everyone that they need to re-verify. They will then keep their grade but will need to authenticate with FEIDE to register for events.
- Finish the grade selection code and remove everyone from their grade. This will mean no changes anywhere else and will ensure _everyone_ will need to re-verify to get their student status back. The downside is that this might be a bit more to implement.

LudvigHz:
> Unique FEIDE users

> Right now, there is no check for the same FEIDE account on multiple users. This means that a single FEIDE user can be used to register several users. This would be quite simple to fix by saving the user ID and un-verifying users if they use the same FEIDE account on another user f.ex.

I have now implemented it in such a way that the `student_username` already used for verification is used again. This time it is retrieved from feide and set on the user. This means that if another user has already taken that username, the verification will fail.

lego/apps/users/models.py
```diff
@@ -441,8 +441,32 @@ def email_address(self):
...
         self.picture = value
 
+    def verify_student(self, feide_groups) -> bool:
+        if self.is_verified_student():
+            return True
+
+        if self.has_grade_group:
+            self.student_verification_status = True
+            self.save()
+            return True
```
eikhr, line 451: Does this mean anyone who is already verified and in a grade group today, can verify without any valid FEIDE group?
I feel like we should save who is verified with FEIDE, and who is verified with the old system in case we want to verify everyone through FEIDE later.

eikhr (approved):
lgtm

lego/apps/users/models.py
LudvigHz: Yes. This is what we discussed on slack and the solution I've implemented here seemed to be the consensus.

About differentiating the two, I see your point and thought about it. But I feel like we would just add extra logic to replace later. Sure, it would be extra work for the users who have authenticated already if we require everyone to re-verify. But it literally takes two seconds, so I don't _really_ feel like it's that big of a deal? The only thing is that anyone who might have been manually verified will need to do that again. But that would just require us to keep a list of those that will apply to.

I am open to suggestions though. But I'd like the solution to not include too many (rather 0) changes to the DB and lots of logic (event registrations etc.)

lego/apps/users/tests/test_student_confirmation_api.py
```diff
@@ -1,103 +1,183 @@
...
+    }
+]
+
+komtek_master_resp = [
+    {
+        "id": "fc:fs:fs:prg:ntnu.no:mstcnns",
+        "type": "fc:fs:prg",
+        "displayName": "Digital Infrastructure and Cyber Security",
+    }
+]
```
PeterJFB, line 91: I don't see this in any of the tests, yeet or use it:)

LudvigHz: Oppsie, missed this. Thanks!

lego/apps/users/models.py
eikhr: Ok. I guess that's fine. We'll just keep a list of the manually verified users

PR #3259, opened 2023-03-23
Merge data from "created_by" to "authors" field

jonasdeluna:
Migrate data from the old created_by field to the new authors field holding multiple authors instead of one.


lego/apps/articles/fixtures/development_articles.yaml
```diff
@@ -36,6 +36,7 @@
       - lorem
       - ipsum
       - sample-text
+    created_by_id: 1
```
LudvigHz, line 39: Add the authors field here as well

lego/apps/articles/migrations/0012_migrate_authors.py
```diff
@@ -0,0 +1,24 @@
...
+from django.conf import settings
+from django.db import migrations
+
+
+def merge_authors(apps, schema_editor):
+    Article = apps.get_model("articles", "Article")
+    for e in Article.objects.all():
+        if e.created_by != None:
+            e.authors.add(e.created_by)
+            e.save()
```
LudvigHz, line 12: I just realized this doesn't make sense when `authors` is not empty. Might be an idea to check if `authors` is empty as well :) (In theory this should not happen, but it's a bad idea to depend on things going the way you assume)

LudvigHz (commented):
Nice :ok_hand: 

Fix up the comments (and preferably also change the commit message to something that makes more sense) and we're good to go :100:

lego/apps/articles/migrations/0012_migrate_authors.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 4.0.9 on 2023-03-16 13:13
+
+from django.conf import settings
+from django.db import migrations
+
+
+def merge_authors(apps, schema_editor):
+    Article = apps.get_model("articles", "Article")
+    for e in Article.objects.all():
+        if e.created_by != None and e.authors == None:
```
eikhr, line 10: Should't this be `e.authors != None`, or can you do `.add()` anyways?

jonasdeluna: It seemed to run fine for me. From testing in shell, the field is None if there's no entries and adding still works.

LudvigHz (approved):

PR #3260, opened 2023-03-23
[Events only] Add slugs to API and support lookup by slug

ollfkaih:
I don't have a clue if this is a *nice* way of doing this, but it works. Feedback appreciated before I look at meetings/joblistings etc. as I expect that to be rather similar to this.

Also see frontend: https://github.com/webkom/lego-webapp/pull/3711

### DIFF
--- a/lego/apps/events/serializers/events.py
+++ b/lego/apps/events/serializers/events.py
@@ -130,6 +130,7 @@ class EventReadDetailedSerializer(
         fields = ObjectPermissionsSerializerMixin.Meta.fields + (
             "id",
             "title",
+            "slug",
             "description",
             "cover",
             "cover_placeholder",
--- a/lego/apps/events/views.py
+++ b/lego/apps/events/views.py
@@ -1,5 +1,7 @@
 from django.db import transaction
 from django.db.models import Count, Prefetch, Q
+from django.http import Http404
+from django.shortcuts import get_object_or_404
 from django.utils import timezone
 from django_filters.rest_framework import DjangoFilterBackend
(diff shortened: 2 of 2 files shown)


lego/apps/content/models.py
```diff
@@ -25,7 +25,7 @@ class Meta:
 
     def generate_slug(self):
```
norbye, line 26: *Not related to this code specifically, I just put it here so I could have a thread(:*

Has there been any discussion about the advantages of this approach vs just adding a "slug-ish" version of the event title to the URL? Such as `https://abakus.no/events/3337` -> `https://abakus.no/events/3337-abacup` or `https://abakus.no/events/3337/abacup` where the slug is ignored and only the id is queried?

Personally I'm partial towards the latter approach, as the slugs don't become outdated if the organizers decide to change the name of the event + we don't end up with dead links if we change the slug later because the id is dominant.

ollfkaih: There is some discussion on [ABA-210](https://linear.app/abakus-webkom/issue/ABA-210/add-slug-to-events-joblistings-interestgroups).
I started out with /{id}/{slug} and just querying the id, but just the slug is prettier I think. 
We could overengineer this properly by changing the slug when title is changed, and also storing all the slugs in the db for query, but this might have performance issues. 

Also, I don't know quite how often event names are changed, but we could just try this approach and if people complain about outdated slugs/they become prevalent, we can address the issue then. At least if slugs never changed after creation, there will be no dead links.

LudvigHz (approved):

PR #3264, opened 2023-03-27
Fix migration of created_by -> authors field properly

jonasdeluna:
Fix migration of created_by -> authors field properly.


ivarnakken (commented):
👍🏼

PR #3299, opened 2023-04-17
Fix tox mypy installation

LudvigHz:
Mypy needs all deps to check types.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,7 +6,10 @@ skipsdist = True
 basepython = python3.11
 allowlist_externals = poetry
 commands_pre =
-    poetry install --only mypy,formatting,flake8
+    isort: poetry install --only formatting
+    flake8: poetry install --only flake8
+    black: poetry install --only formatting
+    mypy: poetry install --with prod --without formatting,flake8
 commands =
     isort: isort -c lego
     flake8: flake8


tox.ini
```diff
@@ -6,7 +6,10 @@ skipsdist = True
 basepython = python3.11
 allowlist_externals = poetry
 commands_pre =
-    poetry install --only mypy,formatting,flake8
+    isort: poetry install --only formatting
+    flake8: poetry install --only flake8
+    black: poetry install --only formatting
+    mypy: poetry install --with prod --without formatting,flake8
```
ivarnakken, line 12:
```suggestion
    mypy: poetry install --only mypy
```

ivarnakken (commented):
`mypy` seems to run fine through `tox` for me? Why do you mean there's a bug?

Are you saying `mypy` will wrongly check files if it does not have access to the deps - makes sense. I suppose it works for me because I already have the deps in the venv. But, you can't run `mypy` through `tox` without `tox` though, so you kind of _have_ to install the default deps first. How have you encountered this?

LudvigHz:
> mypy seems to run fine through tox for me? Why do you mean there's a bug?

What? It's not working at all. It's also failing in all CI runs?

> Are you saying mypy will wrongly check files if it does not have access to the deps - makes sense. I suppose it works for me because I already have the deps in the venv. But, you can't run mypy through tox without tox though, so you kind of have to install the default deps first. How have you encountered this?

Well, mypy needs all deps in the project, because they contain the types? So if tox can't resolve package imports it can't check the types. I don't understand how it could work without installing any dependencies.

You can't have access to the venv, cause tox should create an isolated env with the deps. So even more strange if it works for you :thinking:.

TLDR; As it now it shouldn't work for anyone? It certainly does not work for me or in CI.

ivarnakken (approved):
python venvs is a mess 🤷🏼

PR #3301, opened 2023-04-17
Create a membership history when updating a membership

ivarnakken:
Frontend change: https://github.com/webkom/lego-webapp/pull/3805

Tested manually that the creation of membership history works, but I'll write up a quick test for it soon.

Related to ABA-400

### DIFF
--- a/lego/apps/users/views/memberships.py
+++ b/lego/apps/users/views/memberships.py
@@ -1,3 +1,8 @@
+from datetime import datetime
+from typing import Any
+
+from django.http import HttpRequest
+from django.utils import timezone
 from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework import filters, status, viewsets
 from rest_framework.response import Response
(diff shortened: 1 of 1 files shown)


lego/apps/users/views/memberships.py
```diff
@@ -1,10 +1,16 @@
+from datetime import datetime
```
LudvigHz, line 1: For consistency, use `django.utils.timzone`

lego/apps/users/views/memberships.py
```diff
@@ -33,3 +39,21 @@ def get_queryset(self):
...
         return super(MembershipViewSet, self).create(request, *args, **kwargs)
+
+    def update(self, request: HttpRequest, *args: Any, **kwargs: Any) -> Response:
+        """Create a new membership history object when a membership is updated"""
+
+        user = User.objects.get(pk=request.data["membership"]["id"])
+        group = AbakusGroup.objects.get(pk=request.data["membership"]["abakus_group"])
+        start_date = datetime.strptime(
+            request.data["membership"]["created_at"], "%Y-%m-%dT%H:%M:%S.%fZ"
+        )
```
LudvigHz, line 50: Do `serializer.is_valid()`  and use the serialized value in the `create` below and you can throw away this.

LudvigHz (changes requested):

PR #3312, opened 2023-04-18
Implement Image Gallery Functionality & Endpoints

jonasdeluna:
Implement imagegallery endpoints


lego/apps/events/serializers/events.py
```diff
@@ -192,6 +192,23 @@ def get_attended_count(self, event):
         return event.registrations.filter(presence=PRESENCE_CHOICES.PRESENT).count()
 
 
+class ImageGallerySerializer(BasisModelSerializer):
+    cover = ImageField(source="key", required=True, options={"height": 500})
+
```
LudvigHz, line 197: As with the image fields in other serializers, I would suggest also adding a `placeholder` that can plug into the image component on the frontend for smoother loading. You can just copy what's done in the serializers over here :).

lego/apps/events/views.py
```diff
@@ -297,6 +299,19 @@ def upcoming(self, request):
         serializer = self.get_serializer(queryset, many=True)
         return Response(serializer.data)
 
+    @decorators.action(
+        detail=False,
+        serializer_class=ImageGallerySerializer,
+        permission_classes=[permissions.IsAuthenticated],
```
LudvigHz, line 305: Hmm, I think we should discuss the permissions here. Might be a good idea to use event create permissions for this?

lego/apps/files/serializers.py
```diff
@@ -1,8 +1,43 @@
...
 
 class FileUploadSerializer(serializers.Serializer):
     key = serializers.RegexField(regex=KEY_REGEX)
     public = serializers.BooleanField(required=True)
+
+
+class ImageGalleryCoverSerializer(serializers.ModelSerializer):
+    file = ImageField(
+        source="self", required=False, options={"height": 700, "smart": True}
+    )
```
LudvigHz, line 17: This is unused?

lego/apps/files/views.py
```diff
@@ -46,6 +46,21 @@ def create(self, request, *args, **kwargs):
...
+        methods=["PATCH"],
+        permission_classes=[permissions.AllowAny],
+    )
+    def set_save_for_use(self, request, *args, **kwargs):
+        file = self.get_object()
+        serializer = FileSaveForUseSerializer(data=request.data)
+        if serializer.is_valid():
+            file.save_for_use = serializer.validated_data["save_for_use"]
+            file.save()
+            return Response({"Success"})
```
LudvigHz, line 60: Avoid using arbitrary return values here and rather just return a `HTTP_200` f.ex.

start.sh
```diff
@@ -0,0 +1,4 @@
+docker compose up -d
```
LudvigHz, line 1: Remove

lego/apps/files/serializers.py
```diff
@@ -1,8 +1,43 @@
...
+
 from .validators import KEY_REGEX
 
 
 class FileUploadSerializer(serializers.Serializer):
     key = serializers.RegexField(regex=KEY_REGEX)
     public = serializers.BooleanField(required=True)
+
+
+class ImageGalleryCoverSerializer(serializers.ModelSerializer):
```
LudvigHz, line 14: Where is this serializer used?

lego/apps/files/views.py
```diff
@@ -46,6 +46,21 @@ def create(self, request, *args, **kwargs):
             status=status.HTTP_201_CREATED,
         )
 
+    @decorators.action(
+        detail=True,
+        methods=["PATCH"],
+        permission_classes=[permissions.AllowAny],
+    )
```
LudvigHz, line 53: This definitely needs other permissions :))

I realize this might actually be a little tricky, but here are my thoughts:
- Have permission class set to authenticated f.ex.
- Do a permission check on the object. Since I don't believe we have a permission class for `File`s?? you could just do the permission check inline here in the view (or create a one-off function for it). I think you'd need to get the related object of the file (`Article`, `Event` (maybe just do `Content` actually)) and do a permission check on that object.

The _easier_ way would be to not allow this patch at all and just set this flag on creating the event/article. What is actually the reasoning for having this endpoint?

LudvigHz: When I read some more of the `FileField` code, I see that much of this _might_ be covered there? Could you confirm that? If that's the case you could also use the logic that exists there and apply it here. You could also just use the `FileField` directly to simplify things I guess, even though that might be a bit weird.

If you could write up some tests to back up any hypothesis/claims on how the permission checks work here that would be fantastic

lego/apps/events/constants.py
```diff
@@ -10,6 +10,7 @@
 SOCIAL = "social"
 OTHER = "other"
 EVENT = "event"
+IMAGE = "image"
```
LudvigHz, line 13: What's this used for?

LudvigHz (changes requested):
Super nice. this looks very good. Just some questions that need addressing and some discussions to be had :)

Also, the naming could be better IMO :grimacing:. (ImageGallery is very vague, and something like `CoverImageGallery` for the endpoint makes more sense)

lego/apps/files/serializers.py
jonasdeluna: Removed, unused.

lego/apps/events/constants.py
jonasdeluna: Already defined somewhere else, removed and fixed import.

lego/apps/files/views.py
jonasdeluna: Added EDIT event permission check

lego/apps/files/views.py
```diff
@@ -46,6 +49,23 @@ def create(self, request, *args, **kwargs):
...
+    def set_save_for_use(self, request, *args, **kwargs):
+        if request.user.has_perm(EDIT, Event) is False:
+            raise PermissionDenied()
+        file = self.get_object()
+        serializer = FileSaveForUseSerializer(data=request.data)
+        if serializer.is_valid():
+            file.save_for_use = serializer.validated_data["save_for_use"]
+            file.save()
+            return Response(data=serializer.data, status=status.HTTP_200_OK)
```
LudvigHz, line 65: I can't see that you ever check permissions on the image here?

I can elaborate: If you have event EDIT perm, then it looks to me like you can set this variable on any image as long as you have the file key. I.e. you can access an image you should not be able to access as long as you have the key and event EDIT perm. 

You need to implement some sort of perm check here. Even though it's crude, I think using the token auth as in the `FileField` should be adequate, as that token should never be shared.

LudvigHz (changes requested):
Still some permission stuff missing, but then this should be good to go!

Would also be nice if you could add some tests for the endpoints you created here :)
(check different perms, and that the response is correct in each case)

lego/apps/files/views.py
```diff
@@ -46,6 +49,32 @@ def create(self, request, *args, **kwargs):
...
+                raise PermissionDenied()
+            try:
+                file.save_for_use = serializer.validated_data["save_for_use"]
+                file.save()
+            except KeyError:
+                return Response(
+                    {"This field is required!"}, status=status.HTTP_400_BAD_REQUEST
+                )
+            except Exception:
+                return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
LudvigHz, line 73: Don't except `Exception` like this. Either catch the specific error or let it throw.

lego/apps/files/views.py
```diff
@@ -46,6 +49,32 @@ def create(self, request, *args, **kwargs):
...
+        if serializer.is_valid():
+            if file.token != serializer.validated_data["token"]:
+                raise PermissionDenied()
+            try:
+                file.save_for_use = serializer.validated_data["save_for_use"]
+                file.save()
+            except KeyError:
+                return Response(
+                    {"This field is required!"}, status=status.HTTP_400_BAD_REQUEST
+                )
```
LudvigHz, line 71: How and when does this exception occur? If you are missing a field, this should be handled by the serializer. (The `serializer.is_valid()` will throw a validation error when you're missing a field or the field validation fails there).

lego/apps/files/tests/test_api.py
```diff
@@ -44,3 +44,128 @@ def test_post_create_file_call(self, mock_create_file):
...
+
+class SetSaveForUserTest(BaseAPITestCase):
+    fixtures = ["test_users.yaml", "test_files.yaml", "test_abakus_groups.yaml"]
+    key = "abakus.png"
+    token = "token"
+    url = f"/api/v1/files/{key}/set_save_for_use/"
+
+    def setUp(self):
+        self.webkom_user = User.objects.get(pk=9)
+        self.random_user = User.objects.get(pk=1)
```
LudvigHz, line 57: What does `random_user` mean? Is it a normal user? If so preferable use a name that reflects the object :)

lego/apps/files/tests/test_api.py
```diff
@@ -44,3 +44,128 @@ def test_post_create_file_call(self, mock_create_file):
         except Exception:
             pass
         mock_create_file.assert_called_with(key, self.user, False)
+
+
+class SetSaveForUserTest(BaseAPITestCase):
+    fixtures = ["test_users.yaml", "test_files.yaml", "test_abakus_groups.yaml"]
+    key = "abakus.png"
+    token = "token"
+    url = f"/api/v1/files/{key}/set_save_for_use/"
```
LudvigHz, line 53: Use the `reverse` method instead of using url strings directly

LudvigHz: Also, if you could come up with a better endpoint for this that would be super nice. I don't think `set_save_for_use` is specifically meaningful or nice.

LudvigHz (commented):
Awesome with the tests!

Just a few comments to clean things up and this should be good to go :)

lego/apps/files/serializers.py
```diff
@@ -1,8 +1,29 @@
...
+        fields = (
+            "token",
+            "save_for_use",
+        )
+
+    def validate(self, data):
+        if self.instance.token != data["token"]:
+            raise PermissionDenied()
+        if "save_for_use" not in data:
+            raise ValidationError()
```
LudvigHz, line 28: Would be nice to include an error message. but not critical

lego/apps/files/views.py
```diff
@@ -46,6 +49,22 @@ def create(self, request, *args, **kwargs):
...
+    def imagegallery(self, request, *args, **kwargs):
+        if request.user.has_perm(EDIT, Event) is False:
+            raise PermissionDenied()
+        file = self.get_object()
+        serializer = FileSaveForUseSerializer(file, data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(data=serializer.data, status=status.HTTP_200_OK)
+        else:
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
LudvigHz, line 66: Since you have `raise_exception` in `is_valid()`, you won't execute the code path in the else block no?

Not an issue though :)

lego/apps/files/tests/test_api.py
```diff
@@ -44,3 +45,132 @@ def test_post_create_file_call(self, mock_create_file):
         except Exception:
             pass
         mock_create_file.assert_called_with(key, self.user, False)
+
+
+def _get_detail_url(key):
+    return reverse("api:v1:files-imagegallery", kwargs={"key": key})
+
+
+class SetSaveForUserTest(BaseAPITestCase):
```
LudvigHz, line 54: This is a super clean and thorough test! Love it!! :100:

LudvigHz (approved):
Awesome!

PR #3313, opened 2023-04-19
Fix docker build with poetry

eikhr:
Changed `COPY requirements /app/requirements` to `COPY pyproject.toml /app/pyproject.toml` in Dockerfile

`requirements` no longer exists, causing the build to fail. `pyproject.toml` now contains the requirements, and is needed to install deps with poetry.


LudvigHz (approved):
Nice one :upside_down_face:

ivarnakken (approved):
Ooops

PR #3321, opened 2023-04-19
Make ical feed generation of meetings not slow

LudvigHz:
The lookup for participating was very expensive and slow. It would cause
timeouts so some users' calendars would not load.

~~UNTESTED~~

~~This is my theory at least. I just wrote this code and haven't had the time to test it. But events
does not suffer from this problem so I hope my suspicions are correct.~~

Tested the code in staging with a hotpatch. Works great now :)


LudvigHz:
Sorry about the noise in the diff. I also fixed some types here and there when I was working :))

ollfkaih (approved):
LGTM🥃

eikhr (approved):
LGTM, much needed!

PR #3334, opened 2023-04-26
Fetch event analytics from Plausible

ivarnakken:
It's own decorator used by the webapp on /statistic.

[Related frontend PR.](https://github.com/webkom/lego-webapp/pull/3840)


norbye (approved):

lego/apps/events/views.py
```diff
@@ -250,6 +255,30 @@ def allergies(self, request, *args, **kwargs):
...
+
+        headers = {
+            "Authorization": f"Bearer {settings.PLAUSIBLE_KEY}",
+            "Content-Type": "application/json",
+        }
+
+        created_at = event.created_at.strftime("%Y-%m-%d")
+        now = datetime.now().strftime("%Y-%m-%d")
+        # Plausible wants the date on this schema: YYYY-MM-DD,YYYY-MM-DD
+        date = f"{created_at},{now}"
```
LudvigHz, line 271: How will this affect events created a long time ago? (In terms of amount of data etc)

ivarnakken: Ye that's a problem, but I'm planning on giving the user the ability to filter on different timeseries. This is still a nice mvp though, since we don't have plausible analytics for old events.

PR #3340, opened 2023-04-27
Fix DAPHNE_SERVER setting check

LudvigHz:
The first arg could be "/usr/local/daphne" f.ex. This should not break
anything as the first arg would always be the path to the daphne
executable.


ivarnakken (approved):
:shipit:

PR #3370, opened 2023-05-05
Add slugs to joblistings and articles

ollfkaih:
- Add slugs to joblistings
- Add slugs to articles
- Add slugs to a couple missing views in events 

Resolves ABA-211


ivarnakken (approved):
Looks good 👍🏼 

---

Not sure if I get your fix tests commits.

ollfkaih:
> Not sure if I get your fix tests commits.

They changed the code so the tests would pass (mostly returning specified error codes)

PR #3389, opened 2023-05-13
GitHub profile username

jonasdeluna:
Implement github username field on the user model.


ivarnakken (approved):
:octocat:

lego/apps/users/validators.py
```diff
@@ -17,6 +17,13 @@
     "invalid",
 )
 
+github_username_validator = validators.RegexValidator(
+    r"^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}$",
```
LudvigHz, line 21: What is this regex/where did you get it from?

jonasdeluna: Just stole it from here: https://github.com/shinnn/github-username-regex

PR #3450, opened 2023-07-03
Move `mypy` conf for it to work as intended

ivarnakken:
The quotes around the regex was originally added to comply with TOML linting, but this turned out to break the configuration. Therefore the conf had to be moved away from the TOML file.

After moving it, it appeared we were missing a required section for the `django-stubs` plugin, so that was also included.


norbye (approved):
Can't say I've read a lot into mypy configuration - but if it works it looks good enough for me(:

PR #3461, opened 2023-07-13
Allow unauthenticated users to view group details

ivarnakken:


LudvigHz (commented):
I know the perm system is overly conplex. But it might be better to use the `safe_perms` property (or something like that). I _think_ it's on the handler class. But a quick grep should lead you to it.

ivarnakken:
> I know the perm system is overly conplex. But it might be better to use the safe_perms property (or something like that). I think it's on the handler class. But a quick grep should lead you to it.

There is nothing called `safe_perms`, but are you referring to `safe_methods` in `PermissionHandler`? But the `safe_methods` attribute by itself does not grant permissions; it merely defines which actions are considered "safe" or non-modifying. The actual permission checks are performed in methods such as `has_perm` and `has_object_permissions`? Or am I missing something?

LudvigHz:
> There is nothing called `safe_perms`, but are you referring to `safe_methods` in `PermissionHandler`? If so, the `safe_methods` attribute by itself does not grant permissions; it merely defines which actions are considered "safe" or non-modifying. The actual permission checks are performed in methods such as `has_perm` and `has_object_permissions`? Or am I missing something?

Yes, that's the one. The default permission handler will check this attribute against the action and allow the method accordingly. So if you set it to `view` it should allow users to view the objects in the viewset. Although it requires that the group perm handler does not override too much of the behavior already (it really shouldn't)

ivarnakken:
> Yes, that's the one. The default permission handler will check this attribute against the action and allow the method accordingly. So if you set it to view it should allow users to view the objects in the viewset. Although it requires that the group perm handler does not override too much of the behavior already (it really shouldn't)

ah yes I now see that `PermissionHandler`'s has a similar check for this in `has_perm`
```py
        if require_auth and not authenticated:
            return False
        elif not require_auth and perm in self.safe_methods:         # <----------
            return True
        if not authenticated:
            return False
```

Adding `authentication_map = {VIEW: False, LIST: False}` sets `require_auth` to `False` on `VIEW` and `LIST` methods.

Seems to work fine.

LudvigHz (approved):
Is this all groups now??

I.e. can any person view any group?

ivarnakken:
> Is this all groups now??
> 
> I.e. can any person view any group?

Yes, but not its members. I don’t really see anything wrong with letting them be public, but I’m all ears

LudvigHz:
The only concern was that there are some random groups that are administrative only and not really meant to be accessible by the public. But yeah at least it's not the members but still we might want to have some filter on what is public (in terms of group type (committee, interest group))

PR #3471, opened 2023-08-13
Fix bad naming of MSTCNNS

ivarnakken:
I suppose it needs to be uppercase like the rest? Mock tests still pass, but there's no way for me to actually test this ..

Again, I have no idea if this actually fixes the bug 🙃 Anyone got any ideas?


eikhr (approved):
Yea, sikkert det som er problemet

LudvigHz (changes requested):
Aaaah okey right but please change the code to be case invariant

PR #3472, opened 2023-08-13
Rename studies back to uppercase

ivarnakken:
We missed a single .lower(), so instead of having to add that every time you want to do something I think it's better to just keep them as uppercase, especially considering we get them like that from Feide.

image

### DIFF
--- a/lego/apps/users/constants.py
+++ b/lego/apps/users/constants.py
@@ -80,10 +80,10 @@ class FSGroup(Enum):
     def values(cls) -> list[str]:
         return [e.value for e in cls]
 
-    MTDT = "fc:fs:fs:prg:ntnu.no:mtdt"
-    MTKOM = "fc:fs:fs:prg:ntnu.no:mtkom"
-    MIDT = "fc:fs:fs:prg:ntnu.no:midt"
-    MSTCNNS = "fc:fs:fs:prg:ntnu.no:mstcnns"
+    MTDT = "fc:fs:fs:prg:ntnu.no:MTDT"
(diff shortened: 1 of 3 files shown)


LudvigHz (approved):
LGTM
Let's hope this is consistent from FEIDE and FSWEB :crossed_fingers:

lego/apps/users/constants.py
```diff
@@ -80,10 +80,10 @@ class FSGroup(Enum):
...
         return [e.value for e in cls]
 
-    MTDT = "fc:fs:fs:prg:ntnu.no:mtdt"
-    MTKOM = "fc:fs:fs:prg:ntnu.no:mtkom"
-    MIDT = "fc:fs:fs:prg:ntnu.no:midt"
-    MSTCNNS = "fc:fs:fs:prg:ntnu.no:mstcnns"
+    MTDT = "fc:fs:fs:prg:ntnu.no:MTDT"
+    MTKOM = "fc:fs:fs:prg:ntnu.no:MTKOM"
+    MIDT = "fc:fs:fs:prg:ntnu.no:MIDT"
+    MSTCNNS = "fc:fs:fs:prg:ntnu.no:MSTCNNS"
```
PeterJFB, line 87: As long as we are 100% certain that these are correctly formatted :eyes:

PeterJFB (approved):
Lgtm :+1:, but

lego/apps/users/constants.py
ivarnakken: Time will tell

PR #3486, opened 2023-09-06
Remove report from search serializer

jonasdeluna:
In the future we may want to allow searching for meeting reports you are a part of but this is a quick fix for the security issue.

Resolves: ABA-563

### DIFF
--- a/lego/apps/meetings/search_indexes.py
+++ b/lego/apps/meetings/search_indexes.py
@@ -10,13 +10,12 @@ class MeetingModelIndex(SearchIndex):
     result_fields = (
         "title",
         "description",
-        "report",
         "start_time",
     )
     autocomplete_result_fields = ("title", "start_time")
 
     autocomplete_fields = ("title", "description")
-    search_fields = ("title", "description", "report")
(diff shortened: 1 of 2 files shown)


lego/apps/meetings/search_indexes.py
```diff
@@ -10,13 +10,12 @@ class MeetingModelIndex(SearchIndex):
     result_fields = (
```
norbye, line 10: Have you tried changing the `queryset` to something along the lines of `Meeting.objects.filter(invited__includes=user.id)` (or some other variation that might be more correct:p)?

norbye (approved):
Would be awesome if you consider playing with the queryset before going for this approach - but I'll leave it as approved so it's up to you

norbye:
Whops just noticed that this was set to canceled in linear - so maybe not heheh - just close the PR if it's not happening

danielyanghansen:
As per [ABA-563](https://linear.app/abakus-webkom/issue/ABA-563) and what we discuesed on our meeting like a month ago, this can be closed

PR #3489, opened 2023-09-12
Adds user's name to reactionsGrouped for meetings

jonasdeluna:


lego/apps/meetings/models.py
```diff
@@ -52,6 +52,7 @@ def get_reactions_grouped(self, user):
                     "count": 0,
                     "has_reacted": False,
                     "reaction_id": None,
+                    "user": user.full_name,
```
eikhr, line 55: I don't think this is the user you want. This will always be the currently logged in user.

eikhr (commented):
I think we will need to change a bit more than this, as the reactions_grouped field might contain reactions from multiple different users. 

It might be easier to just use the reactions individually instead of grouping them, if it's not too slow. There used to be some code for this in redux, but it might have been removed a while a ago.

PR #3495, opened 2023-09-27
Give permission to see Allergies for Event to Responsible Group

danielyanghansen:
Backend checks if user is part of responsible group when delegating authorization to view allergies.

The corresponding serializer has also been updated to include the "ResponsibleGroup" field so that frontend also can check if user should be able to see the allergy tab

Resolves: ABA-589

### DIFF
--- a/lego/apps/events/models.py
+++ b/lego/apps/events/models.py
@@ -32,7 +32,7 @@ from lego.apps.files.models import FileField
 from lego.apps.followers.models import FollowEvent
 from lego.apps.permissions.models import ObjectPermissionsModel
 from lego.apps.users.constants import AUTUMN, SPRING
-from lego.apps.users.models import AbakusGroup, Penalty, User
+from lego.apps.users.models import AbakusGroup, Membership, Penalty, User
 from lego.utils.models import BasisModel
 from lego.utils.youtube_validator import youtube_validator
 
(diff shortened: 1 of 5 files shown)


lego/apps/events/models.py
```diff
@@ -132,6 +132,17 @@ def user_should_see_regs(self, user: User) -> bool:
             or user.is_abakom_member
             or (self.created_by is not None and self.created_by.id == user.id)
         )
+    
+    def user_should_see_allergies(self, user: User) -> bool:
+        memberships = Membership.objects.filter(user=user)
+        in_responsible_group = self.responsible_group in [mem.abakus_group for mem in memberships]
```
ivarnakken, line 138: Is it not better to just check if the user is part of the responsible group than to go through all memberships and see if the group is there?

lego/apps/events/models.py
```diff
@@ -132,6 +132,17 @@ def user_should_see_regs(self, user: User) -> bool:
...
+        created_by_self = user == self.created_by
+
+        return (
+            created_by_self
+            or in_responsible_group
+        )
```
ivarnakken, line 144: Nice edge case covered💯

ivarnakken (commented):
Would love to see some unit tests as well😁

Arashfa0301:
Looks good to me :)

Arashfa0301 (approved):

lego/apps/events/models.py
danielyanghansen: I guess? I just assumed that user's have a "friendship paradox" relation to groups. As in: The average group has more members than the average member has groups

itsisak (approved):
LGTM💯

jonasdeluna (approved):

jonasdeluna:
ABA-589

danielyanghansen:
Tests run fine (`tox -e tests`)

PR #3499, opened 2023-10-02
Change bedkom booking period

juniwbjerde:
Change the dates for when automatic answer mails to company interests are sent


eikhr (approved):
I've looked over your changes, and couldn't find any obvious mistakes:)

PR #3501, opened 2023-10-04
Implement lending system

Arashfa0301:
Resolves ABA-613


eikhr (commented):
We need an image field in `lendableObject`, and `responsibleRole` needs to support multiple roles (and should then be called `responsibleRoles`)

lego/api/v1.py
```diff
@@ -210,4 +215,5 @@
 urlpatterns = [
     path("", include(router.urls)),
     path("forums/", include((forums_urls, "forums"))),
+    path("", include((lending_urls, "lending"))),
```
eikhr, line 218: Why do we add this here instead of just registering it with the router above?

lego/apps/lending/fixtures/development_lending_requests.yaml
```diff
@@ -0,0 +1,17 @@
+- fields:
+    lendable_object: 1
+    created_by: 3
+    start_date: '2024-02-02T23:16:00+00:00'
+    end_date: '2024-02-02T23:18:00+00:00'
+    message: "Jeg vil låne grill :)"
+    pending: True # Should be a status field with multiple options
```
eikhr, line 7: Wouldn't it be good to make this an enum from the start? To avoid having to make migrations later.

lego/apps/lending/notifications.py
```diff
@@ -0,0 +1,40 @@
+from lego.apps.lending.models import LendingInstance
+from lego.apps.notifications.constants import LENDING_INSTANCE_CREATION
+from lego.apps.notifications.notification import Notification
+from lego.apps.users.models import User
+
+
+class LendingInstanceNotification(Notification):
```
eikhr, line 7: Could we make a more descriptive name for this. I assume it is sent to the responsible group when a new lending request is created?
I think this will be confusing if we add more notifications later, f.ex. status changes like when the lending instance is approved/denied.

lego/apps/lending/managers.py
```diff
@@ -0,0 +1,44 @@
+from lego.utils.managers import BasisModelManager
+
+
+class LendingInstanceManager(BasisModelManager):
+    def create(self, *args, **kwargs):
+        from lego.apps.lending.notifications import LendingInstanceNotification
+        from lego.apps.users.models import Membership, User
```
eikhr, line 7: Idk how python works, is it really a good idea to import these inside the function?

lego/apps/lending/managers.py
```diff
@@ -0,0 +1,44 @@
...
+            user = User.objects.get(pk=user_id)
+            notification = LendingInstanceNotification(
+                lending_instance=lending_instance,
+                user=user,
+            )
+            notification.notify()
+
+        return lending_instance
+
+    # TODO: We probably need another template for accepting instances
```
eikhr, line 30: I don't understand what it means, but remember to do it or remove the TODO.

lego/apps/lending/managers.py
```diff
@@ -0,0 +1,44 @@
...
+    def update(self, *args, **kwargs):
+        from lego.apps.lending.notifications import LendingInstanceNotification
+
+        lending_instance = super().update(*args, **kwargs)
+        notification = LendingInstanceNotification(
+            lending_instance=lending_instance,
+            user=lending_instance.created_by,
+        )
+        notification.notify()
```
eikhr, line 39: This must be wrong? This notification says that a new lending request was created, right? And it is sent to the user that created it when it is updated??

lego/apps/lending/migrations/0005_alter_lendableobject_responsible_roles.py
eikhr, line 1: I believe it is best practice to remove all these separate migrations, and just create one initial migration as there is no data in prod yet that needs these.

lego/apps/lending/managers.py
jonasdeluna: It's like a lazy import to avoid circular dependencies, I think it's done other places too

eikhr: Ah, that makes sense:)

jonasdeluna:
@eikhr I'll look closer at the comments after the 2 upcoming exams 👍👍

lego/api/v1.py
jonasdeluna: I think it's because the routing structure is a little different than what the viewset would indicate

lego/apps/lending/managers.py
jonasdeluna: Removed it for now.

jonasdeluna:
@eikhr I've fixed your comments, notifications still don't work so they are removed for now. Will squash when merged.

lego/apps/comments/models.py
```diff
@@ -34,5 +36,29 @@ def delete(self):
...
+                    "reaction_id": None,
+                }
+
+            grouped[reaction.emoji.pk]["count"] += 1
+
+            if reaction.created_by == user:
+                grouped[reaction.emoji.pk]["has_reacted"] = True
+                grouped[reaction.emoji.pk]["reaction_id"] = reaction.id
+
+        return sorted(grouped.values(), key=lambda kv: kv["count"], reverse=True)
```
eikhr, line 57: What's going on here? Why is this in this PR?

lego/apps/email/serializers.py
```diff
@@ -44,6 +46,33 @@ class Meta:
             "additional_emails",
         )
 
+    def validate(self, attrs: Any) -> Any:
```
eikhr, line 49: And this stuff has nothing to do with the lending system

eikhr (changes requested):
I think something weird has happened during a rebase maybe? A lot of unrelated changes in the diff

jonasdeluna:
> I think something weird has happened during a rebase maybe? A lot of unrelated changes in the diff

Hmm actually I merged instead of rebasing, how can I resolve it now?

eikhr:
> > I think something weird has happened during a rebase maybe? A lot of unrelated changes in the diff
> 
> Hmm actually I merged instead of rebasing, how can I resolve it now?

I don't know, I haven't seen this before.
Maybe a rebase will help?

PR #3511, opened 2023-10-20
Add revue to the permitted groups for the contact form

haukkagu:
Initially, I wanted to add support for only contacting the revue board. However, the only way to achieve that would be to hardcode in the group id (or handle RevyStyret as a special group like we do Hovedstyret). I decided to just add support for all revue groups, because that was a lot easier. 

Related to [this PR](https://github.com/webkom/lego-webapp/pull/4221).

### DIFF
--- a/lego/apps/contact/serializers.py
+++ b/lego/apps/contact/serializers.py
@@ -1,6 +1,7 @@
+from django.db.models import Q
 from rest_framework import exceptions, serializers
 
-from lego.apps.users.constants import GROUP_COMMITTEE
+from lego.apps.users.constants import GROUP_COMMITTEE, GROUP_REVUE
 from lego.apps.users.models import AbakusGroup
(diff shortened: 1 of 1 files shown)


danielyanghansen (approved):
I actually think it's quite nice to allow all of the revue groups to be contacted. The subgroups have leaders anyway, and groups like the band can deffo be relevant to contact (if they want to be available)

anyways: Looks good

jonasdeluna (approved):
LGTM, is there a frontend change adding the revue to the group picker or is that automatic?

ivarnakken (approved):

haukkagu:
> LGTM, is there a frontend change adding the revue to the group picker or is that automatic?

Yup. [This one](https://github.com/webkom/lego-webapp/pull/4221).

eikhr (approved):

PR #3518, opened 2023-11-03
Enable filtering memberships by multiple roles

norbye:
Memberships can now be filtered by multiple roles - which makes it possible to do the same in the frontend.

Frontend PR: https://github.com/webkom/lego-webapp/pull/4257

The filtering still works the old way, so this should be good to merge regardless of the frontend PR

ABA-366


eikhr (approved):

haukkagu (approved):

danielyanghansen (approved):

Arashfa0301 (approved):
Looks good

PR #3519, opened 2023-11-07
Remove revue and add board to the permitted groups for the contact form

haukkagu:
Now that RevyStyret has has the board group type, this needs to be changed...


juniwbjerde (approved):
People will be able to contact the other boards as well now, but I think that's just fine:)

PR #3529, opened 2023-11-20
Remove user field from survey submission read serializer

eikhr:
The surveys are supposedly anonymous, they are not when the user field is sent with the submission

Related frontend PR: https://github.com/webkom/lego-webapp/pull/4988

TODO:
- [x] Double-check that nothing breaks in lego-webapp, especially for non-admin users

Resolves ABA-273


jonasdeluna (approved):
LGTM :rock:

eikhr:
Turns out this breaks the survey submission page. It uses the user-field to find your own submission

jonasdeluna:
> Turns out this breaks the survey submission page. It uses the user-field to find your own submission

How about leaving it as null unless it's your submission, doing the finding in the backend instead?

ivarnakken (approved):

PR #3550, opened 2024-01-25
Add admissions oauth fixture

norbye:
Makes it so that you don't have to create your own oauth application when devving in admissions - you can just load the fixtures and voilá

image

you can log in with your local LEGO instance


ivarnakken (approved):
Cool I never thought of that!

PR #3349, opened 2023-04-30
Implement new penalty rules

Arashfa0301:
Implement new rules and consequences for penalties. The changes are as following:

- If you have at least one mark, you must wait for 5 hours after the regular registration start to sign up for an event. 
- The duration of each new penalty is 10 days or after being banned from participating in 6 events where you have registration eligibility
- Each new penalty is added to the end of the last active penalty. So if 5 days passes after the first penalty is given, and then a new penalty is given, the remaining time is then 15 days

The biggest change is the addition of PenaltyGroup for the purpose of separating penalty consequence calculation logic and the the information used in front-end. 

I have deleted some tests that believe are not needed for our new implementation and i have also made changes to events app that i deemed necessary. I would appreciate reviewing specially these changes as it might be possible that i made some mistakes.

Resolves ABA-92

PR #3564, opened 2024-03-06
Introduce "too late" as an event presence status

ivarnakken:
Similar to the "not present" status, the user will automatically receive a penalty. I figured all previous penalties of type "presence" can be removed since they should not stack.

Also, this resolves a bug, as it is currently not possible to auto delete previous penalties related to the event due to the `.delete()` not being at the end the atomic block. Not sure why it had to be atomic in the first place.


lego/apps/events/constants.py
```diff
@@ -70,6 +70,7 @@
 class PRESENCE_CHOICES(models.TextChoices):
     UNKNOWN = "UNKNOWN"
     PRESENT = "PRESENT"
+    TOO_LATE = "TOO_LATE"
```
ivarnakken, line 73: I'd like some input on this naming as well - not sure if I like it or not

ivarnakken: I can live with that

lego/apps/events/models.py
```diff
@@ -930,26 +930,46 @@ def set_presence(self, presence: constants.PRESENCE_CHOICES) -> None:
...
+    def delete_penalties_for_event(self) -> None:
+        for penalty in self.user.penalties.filter(source_event=self.event):
+            penalty.delete()
+
     def handle_user_penalty(self, presence: constants.PRESENCE_CHOICES) -> None:
+        """
+        Previous penalties related to the event are deleted since the
+        newest presence is the only one that matters
+        """
+
```
norbye, line 947: Is the logic here that if you get penalties for not paying on time you will no longer have those penalties if you also didn't show up to the event? I would've thought that they rather stack on top of each other - so that each penalty-worthy action got you a new penalty? Or is that not how we want to do it?

lego/apps/events/models.py
```diff
@@ -930,26 +930,46 @@ def set_presence(self, presence: constants.PRESENCE_CHOICES) -> None:
...
+            )
+        elif (
+            self.event.heed_penalties
+            and presence == constants.PRESENCE_CHOICES.TOO_LATE
+        ):
+            self.delete_penalties_for_event()
+            Penalty.objects.create(
+                user=self.user,
+                reason=f"Møtte for sent opp på {self.event.title}.",
+                weight=1,
```
norbye, line 968: Do we want to store this in the event, alongside the penalty_weight and penalty_weight_on_not_present? As far as I see it makes little difference cause we don't seem to allow changing those values in the frontend - but it might be neat to keep them together 🤷 

Or add some constant variable for the different penalty weights in a constants.py file

ivarnakken: I think making it a constant is sufficient - no need to over engineer this

lego/apps/events/models.py
ivarnakken: > I would like some input as to how this will affect other penalties, e.g. given due to late payments (or other things?), as they will also be removed if the presence is changed to something worthy of a penalty. Should we consider to introduce "penalty types" stored on the record, e.g. "presence", "payment", "other"?

ivarnakken: Yes, they should definitely stack. Just not sure how this should be implemented.

norbye: Ah right.. I thought the stacking functionality was already implemented in the redo Arash did last year to make penalties last for x amount of days instead of adding onto each other - or maybe I remember that wrong

ivarnakken: so do you have some input? 😄 If we add penalty types, we can flag them as being either presence or payment related, making it easy to delete the right penalties

norbye: image

Don't have too much rn - but I'll look through the penalty implementation some time after I get back

Initial thoughts were that not too much had to be changed, as penalties already stack (or so I thought), so you could just add the penalty without deleting the others, but by your question it appears that is not the case so I'll have to review it

ivarnakken: haha no worries - have fun!

> Initial thoughts were that not too much had to be changed, as penalties already stack (or so I thought), so you could just add the penalty without deleting the others, but by your question it appears that is not the case so I'll have to review it

I agree that they should stack, most of the time. If both penalties are e.g. related to the event presence, I don't believe they should stack. This logic also works as a safety net by allowing missclicks

norbye: Ah yeah, I very much agree with that! In that way it sounds like a good idea to add types like you suggested, so you could clear the penalties for attendance status for the event when it's changed and keep the penalties for late payment, etc.

norbye: Just making sure that my input is clearly stated now that I'm back on my computer;

I've been thinking about it and I agree that your proposal is probably the best idea (adding penalty types). So then users would get penalties for late payments and such, and when their presence is changed, only the presence-penalties are modified. This would also be a nice way to remove late penalties from users if they get their status changed to present at a later point due to some mistake (or misclicks as you said)

Realised that part of my confusion came from thinking that this pr had been merged ages ago - which it obviously hasn't https://github.com/webkom/lego/pull/3349

lego/apps/users/constants.py
```diff
@@ -142,3 +142,12 @@ def values(cls) -> list[str]:
     (LIGHT_THEME, LIGHT_THEME),
     (DARK_THEME, DARK_THEME),
 )
+
+
+LATE_PRESENCE_PENALTY_WEIGHT = 1
```
norbye, line 147: Would be nice to keep all the penalty weights togheter
```suggestion
class PENALTY_WEIGHTS(models.IntegerChoices):
	PRESENCE = 1
	...
```

norbye (approved):
Awesome stuff!

lego/apps/users/constants.py
```diff
@@ -95,8 +97,6 @@ def values(cls) -> list[str]:
     FSGroup.MSSECCLO: FOURTH_GRADE_KOMTEK,
 }
 
-STUDENT_EMAIL_DOMAIN = "stud.ntnu.no"
-
```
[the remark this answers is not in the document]
ivarnakken: Yes this is not used

lego/apps/users/constants.py
norbye: Seems like maybe you didn't remove the other constant, but that can be fixed later

ivarnakken: oops

PR #3566, opened 2024-03-07
Remove nerd import command

ivarnakken:
It ruins our coverage. Removing this increased our coverage by almost an entire percent


norbye (approved):
Lets goo

eikhr (approved):

PR #3568, opened 2024-03-10
Calculate penalty expiration based on event date instead of current date

haukkagu:
This just makes sense in my opinion. A few incidents have happened where penalty weights have to be changed or a penalty being deleted by accident has to be added again. When the penalty is added again, the expiration date would be later since it is calculated based on when it was created instead of the event date.

**after**:
Screenshot from 2024-03-10 16-34-14

I made these penalties at the same time


haukkagu:
This should probably get fixed when the new penalty rules get merged, so not that important to merge

jonasdeluna (approved):
Nice work 👍👍

ivarnakken (approved):
Yes this makes more sense 👍🏼

Bestem0r (approved):

Bestem0r (approved):
Great change!

PR #3570, opened 2024-03-12
Allow empty notification channels

AdrianAndersen:
Currently, lego throws a nasty error if you try to remove both "email" and "push" from the [notification settings](http://abakus.no/users/me/settings/notifications). This is  because the model for the notifications does not allow an empty list. 

This PR simply allows the list of channels to be empty, which is the expected behaviour.


AdrianAndersen:
Resolves ABA-880

PR #3577, opened 2024-03-20
Extend email-user filter to add boards and ordained

norbye:
Backend PR for https://github.com/webkom/lego-webapp/pull/4576

Create another group type "ordained" to access Ordenen without having to hardcode group id's.

Allow filtering of more groups in the email-list view

Allow groups to be fetched by multiple group types and not only one


norbye:
> 👍🏼 I would love a tiny test as well

mañana mañana

eikhr (approved):

PR #3588, opened 2024-04-12
Set password hash for external sync on newly created users

eikhr:
The reason GSuite sync only works after changing password, is that the password hash used for external sync was not set correctly when creating users. The logic for setting this hash is implemented in the `set_password()` method which is not called when creating a new user. 

I wasn't able to add any logic with access to the new password in the `PasswordHashUser`-model, so I just added a call to `set_password` right after creating new user.

resolves ABA-578

Although this will fix the problem for new users, there is still probably a lot of users that have never changed their password and thus do not have the saved hash. It might be worth adding some warning whenever we try to sync a user with a missing hash.

### DIFF
--- a/lego/apps/users/tests/test_password_change.py
+++ b/lego/apps/users/tests/test_password_change.py
@@ -47,6 +47,7 @@ class TestPasswordChange(BaseAPITestCase):
         self.assertEqual(status.HTTP_400_BAD_REQUEST, response.status_code)
 
     def test_new_password_success(self):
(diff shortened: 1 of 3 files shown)


norbye (approved):
Darn, stole my PR :'( (that I hadn't pushed yet)

Was thinking that it might be an option to temporarily set the hash for users on login if it's not set already, as we have 500 users that currently cannot access confluence

eikhr:
Hmm, yeah that should work. Good idea!

PR #3602, opened 2024-05-29
Save missing password hash on login

norbye:
Not the cleanest code out there but for something that has to piggybank on the token generation to be able to generate the hash properly it felt like one of the most reliable ways of doing it as all the important logic is still performed by the rest_framework_jwt.

Open to other suggestions if anyone has any

Will need to be tested in staging to feel better about it, but works just fine for me locally.

The idea is to leave this in prod for a while to generate some hashes, and when the number of missing hashes is low enough we revert the PR.

Resolves ABA-920


eikhr (approved):
I agree this is an ok way to do it. But maybe add a test?

norbye:
> I agree this is an ok way to do it. But maybe add a test?

Absolutely(:

PR #3633, opened 2024-09-05
Use DRF routing for forum thread routing

eikhr:
Seems to work fine with the normal Django REST Framework routing, like all the other routes use. Saves us some custom code.


norbye (approved):

jonasdeluna (approved):
Nice! Hopefully Ill get around to adding more stuff to the forums so we can remove the GitHub discussion one

PR #3638, opened 2024-09-10
Disallow users to give higher permission then they have

eikhr:
To avoid privilege escalation

Resolves ABA-1059


ivarnakken (approved):

norbye (approved):
Nooo, my loophole 😢

PR #3639, opened 2024-09-16
Disable captcha for stripe event fixture

norbye:
No reason to require captcha for the fixtures that are used by stripe. I guess we could also use a token for the test environment to always accept all captchas like we do in development mode, dont really mind either option


ivarnakken (approved):

norbye:
Whop decided to go for the other option as there are no cases where we would like anything to be blocked by an invalid captcha in testing

ivarnakken (approved):

PR #3642, opened 2024-09-20
Add info for calculating waiting list position within pools

eikhr:
When calculating the waiting list position per pool, we need to know what users from the waiting list are applicable for each one. By sending all groups (including inherited) each user in the waiting list are a member of, we can check who can is applicable for each pool, and thus calculate how many will be bumped to that pool before you.

Related frontend PR: https://github.com/webkom/lego-webapp/pull/4983

ABA-1070


norbye (approved):

Bestem0r (approved):
Looks good to me!

PR #3643, opened 2024-09-25
Remove admin_comment from company

juniwbjerde:
Remove the field admin_comment from the company model. It was never really used...

I also wrote a custom data migration to make a normal comment on the company with the admin_comment (in case any of the admin_comments were actually useful and shouldn't be totally deleted

Corresponding backend changes: https://github.com/webkom/lego-webapp/pull/5001

Resolves ABA-1052


ivarnakken (commented):
Should we totally remove it? I thought we only wanted to remove it from the bdb table since it took up space. I don't think we need to remove it as a feature, but instead move it to a better place (like itDAGENE's admin pages haha)

juniwbjerde:
I talked to Falk about it, and his impression after talking to Bedkom was that they didn't use it. I don't see why we would need both the admin_comment field as well as normal comments on companies. It's just a bit messy in my opinion, and there's nothing you would write in the admin comment that wouldn't fit just as well as a normal comment (in my opinion). And with normal comments you get the date it was written as well, so that you can more easily see if it's updated info or not. Espescially when we don't just delete the data in the admin_comment field completely, but rather mighrated it to a comment, I feel like it is completly fine

Bestem0r:
Welp. I kinda messed up this PR in order to hijack it and get it merged.... 😅

ivarnakken:
I might actually die if this is not squashed

ivarnakken:
Actually, no, do not merge this before fixing the commits.
image

norbye:
I think he recreated it in #3727, so this PR could be closed and abandoned

PR #3645, opened 2024-10-02
Add semester_id query to bdb and make student_contact unique for all semesters

falbru:
- Add semester_id query to bdb
- make student_contact unique for all semesters

image


lego/apps/companies/serializers.py
```diff
@@ -36,6 +38,11 @@ def create(self, validated_data):
         return super().create(validated_data)
 
 
+class StudentCompanyContactSerializer(BasisModelSerializer):
+    class Meta:
+        model = StudentCompanyContact
+        fields = ("id", "company_id", "student_id", "semester_id")
```
jonasdeluna, line 44: Is the semester_id supposed to be the year? (In the example it's "1"?

lego/apps/companies/views.py
```diff
@@ -48,11 +48,17 @@ class AdminCompanyViewSet(AllowedPermissionsMixin, viewsets.ModelViewSet):
...
         .prefetch_related("semester_statuses", "files")
-        .select_related("student_contact")
     )
     pagination_class = None
     permission_handler = CompanyAdminPermissionHandler()
 
+    def _list(self, request, *args, **kwargs):
+        semester_id = request.query_params.get('semester_id', None)
+
+        serializer = self.get_serializer(self.queryset, many=True, context={'semester_id': semester_id})
```
jonasdeluna, line 58: Is there some way to just update the context (so we don't lose request and the other couple things passed down by default)?

jonasdeluna (approved):
Great work! Just wondering if the ID will be the year, I think that makes more sense

lego/apps/companies/views.py
falbru: Good point! I'll look into it

lego/apps/companies/serializers.py
falbru: It's the ID of the Semester object

itsisak (approved):
Looks good! I agree with Jonas, it would be neat using the year instead something like `bdb?semester=spring-2024,` but if we get a decent way to switch between semesters in frontend it also does not really matter.

falbru:
I'll change the query paramter later. Kind of bored of this PR rn and want it merged

PR #3646, opened 2024-10-02
Automatically give penalty when payment is overdue

itsisak:
Currently this checks every day at 0900 if there is any events with overdue payments. If so it gives penalty to the users who have not paid, unregisters them from the event and sends notification. 

There is still some things that I need to fix:

- Check if a user paid after the payment deadline and not just at 0900. As this effectively moves the deadline to 0900 the day after the given payment deadline. But then maybe not unregister them if they actually has paid, only give penalty?
- Test it more.
- Some clean up and better function description

### DIFF
--- a/lego/apps/events/notifications.py
+++ b/lego/apps/events/notifications.py
@@ -8,8 +8,10 @@ from lego.apps.notifications.constants import (
     EVENT_BUMP,
     EVENT_PAYMENT_OVERDUE,
     EVENT_PAYMENT_OVERDUE_CREATOR,
+    EVENT_PAYMENT_OVERDUE_PENALTY,
 )
 from lego.apps.notifications.notification import Notification
+from lego.apps.users.constants import PENALTY_WEIGHTS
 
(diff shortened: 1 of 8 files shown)


itsisak:
`handle_payment_overdue` and similar function names and filenames seem very similar to the already existing `payment_overdue` functionality which is more accurately `payment_imminent` or something similar as it is for notifying users when the payment deadline is near. Im considering renaming everything related to that functionality to make the difference between that and this PR clearer. Thoughts?

falbru (approved):

eikhr (changes requested):
I think a good first iteration to make this mergable without too much work would be the following:
- script runs once a day, giving prikk to every person who has not payed after the deadline
- when someone is awarded a prikk, a flag is set on the registration so that we know they have already gotten the payment prikk
- On the event attendees admin page, this "prikk"-status is shown, so that Arrkom can easily check who has gotten prikk

PR #3652, opened 2024-10-09
Add company contact update timestamp

jonasdeluna:
Is this fine as a property or do we want an actual value to save to DB?

Resolves: ABA-1055


itsisak (approved):
In general for performance it might be better with saving it to the DB when updating, but I think it would be pretty negligible, so I say this is fine.

falbru (commented):
Couldn't this be part of the CompanyContact model instead?

PR #3653, opened 2024-10-09
Fix comment reaction notification preventing event update

ollfkaih:
Small change to ensure that the `user` variable is only set if the "request" key is present in the context, which currently prevents some events from being updated.  

ABA-1106


itsisak (approved):
pretty cool

PR #3665, opened 2024-10-30
Prevent indefinite pending unregister

itsisak:
If `EventHasClosed` exception happens, the registration would stay as pending forever. 

Currently the unregister-button is not displayed correctly when the event has seperate registration and unregistration deadlines so it is possible for this to happen. I am fixing that in frontend so it should not be possible anymore, but the registration should never be able to stay as pending imo anyways.


jonasdeluna (approved):
Awesome!

lego/apps/events/tasks.py
```diff
@@ -145,6 +145,13 @@ def async_unregister(self, registration_id, logger_context=None):
             exception=e,
             registration_id=registration.id,
         )
+        registration.status = constants.FAILURE_UNREGISTER
+        registration.save()
+        notify_user_registration(
+            constants.SOCKET_UNREGISTRATION_FAILURE,
+            registration,
+            error_message="Avregistrering feilet",
```
ivarnakken, line 153: Wouldn't hurt to add why in the error message as well hehe

ivarnakken (approved):
Very nice!!

lego/apps/events/tasks.py
itsisak: Fair, just copied from the other exception

eikhr (approved):

ch0rizo (commented):
LGMT!

PR #3667, opened 2024-10-31
Fix rare race condition in event achievement task 

jonasdeluna:
Fix rare race condition in event task and case where rank achievement was not given correctly.

Resolves: ABA-1160


falbru (approved):
Lgtm

PR #3675, opened 2024-11-04
Prefetch more stuff needed in event detail page

eikhr:
This decreases the number of separate SQL-queries, increasing performance

When testing locally with 11 registered + 10 waiting list users, it decreased number of queries from 114 -> 51

@jonasdeluna I moved `achievements` from `PublicUserWithAbakusGroupsSerializer` to `PublicUserWithGroupsSerializer` to avoid fetching achievements for registrations in events. This shouldn't affect the user detail page, is there any other place we need `achievements`?


ivarnakken (approved):
:shipit:

jonasdeluna:
No currently its only used on the detail page so all good 👍

PR #3677, opened 2024-11-04
Auto remove interest group membership for no longer students

jonasdeluna:
Will work on bump_users, but will not remove already alumni users in interest groups.

Resolves: ABA-982


ivarnakken (approved):
Nice!

PR #3678, opened 2024-11-04
Create db indexes for faster lookup in achievement system

jonasdeluna:
Resolves: ABA-1163


lego/apps/achievements/models.py
```diff
@@ -25,3 +25,14 @@ def percentage(self):
...
         return (achievement_users / total_users) * 100
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["user", "identifier"], name="unique_user_identifier"
+            )
+        ]
+        indexes = [
+            models.Index(fields=["identifier", "user"]),
```
ivarnakken, line 36:
```suggestion
```
Is this redundant? If I've understood this correctly, I believe the unique constraint (`UniqueConstraint(fields=["user", "identifier"])`) automatically creates a unique index on these fields in postgres, and the second index `models.Index(fields=["identifier", "user", "level"])` can already handle queries that only need `identifier` and `user`

ivarnakken (approved):
Awesome! 🐎

lego/apps/achievements/models.py
jonasdeluna: Ah yeah that makes sense! You're right

PR #3679, opened 2024-11-04
Add new penalty achievements

aleksandervestlund:
* Add penalty achievement.
* Will make students in Abakus more likely to avoid penalties.
* Add function to find longest period without receiving a penalty. As per now, the user can go maximum one year between two events in the period for which they receive the achievement.
* Improve `dict` type annotation since the one from `typing` is considered obsolete after 3.10.
* Replace the `Callable`-import to be from `collections.abc`, since the one from `typing` is deprecated.
* Small refactor of `check_event_generic`.
* Dependent on [front-end PR](https://github.com/webkom/lego-webapp/pull/5153).
* 100% not Lego-pin bait...

### DIFF
--- a/lego/apps/achievements/verification.py
+++ b/lego/apps/achievements/verification.py
@@ -37,7 +37,7 @@ def check_poll_responses(user: User, count: int):
 def check_longest_period_without_penalties(user: User, years: int) -> bool:
     days = years * 365
 
-    if not (events := Registration.objects.filter(user=user).order_by("end_time")):
+    if not (events := Registration.objects.filter(user=user).order_by("event__end_time")):
         return False
 
     start_time = events.first().event.end_time
--- a/lego/apps/achievements/verification.py
+++ b/lego/apps/achievements/verification.py
@@ -1,3 +1,5 @@
+import itertools
+
 from django.db.models import Sum
 from django.utils import timezone
 
@@ -34,6 +36,23 @@ def check_poll_responses(user: User, count: int):
 
 def check_total_penalties(user: User, count: int) -> bool:
     return Penalty.objects.filter(user=user).count() >= count
+
+
+def check_longest_period_without_penalties(user: User, days: int) -> bool:
+    penalties = sorted(Penalty.objects.filter(user=user), key=lambda x: x.created_at)
+
+    if not penalties:
+        period = timezone.now() - user.date_joined
+        return period.days >= days
+
+    start_period = penalties[0].created_at - user.date_joined
+    end_period = timezone.now() - penalties[-1].created_at
+    intervals = (
+        second.created_at - first.created_at
+        for first, second in itertools.pairwise(penalties)
+    )
+    max_period = max(start_period, end_period, *intervals)
+    return max_period.days >= days
 
 
 # There is a case where manual payment does not update the payment amount.
--- a/lego/apps/achievements/verification.py
+++ b/lego/apps/achievements/verification.py
@@ -39,7 +39,7 @@ def check_longest_period_without_penalties(user: User, years: int) -> bool:
         return False
     if not (
         events := Registration.objects.filter(
-            user=user, status=SUCCESS_REGISTER
+            user=user, status=SUCCESS_REGISTER, event__end_time__lte=timezone.now()
         ).order_by("event__end_time")
     ):
         return False
--- a/lego/apps/achievements/verification.py
+++ b/lego/apps/achievements/verification.py
@@ -58,7 +58,9 @@ def _generate_penalty_intervals(
 def check_longest_period_without_penalties(user: User, years: int) -> bool:
     if not user.has_grade_group:
         return False
-    if not (registrations := _passed_user_registrations(user).order_by("event__end_time")).exists():
+    if not (
+        registrations := _passed_user_registrations(user).order_by("event__end_time")
+    ).exists():
         return False
 
     days = 365 * years
@@ -79,7 +81,9 @@ def check_longest_period_without_penalties(user: User, years: int) -> bool:
 
             last_registration_in_interval = (
                 _passed_user_registrations(user)
-                .filter(event__end_time__gt=current_end, event__end_time__lte=cutoff_time)
+                .filter(
+                    event__end_time__gt=current_end, event__end_time__lte=cutoff_time
+                )
                 .last()
             )
 
@@ -88,7 +92,9 @@ def check_longest_period_without_penalties(user: User, years: int) -> bool:
                 continue
 
             next_event = (
-                _passed_user_registrations(user).filter(event__end_time__gt=cutoff_time).first()
+                _passed_user_registrations(user)
+                .filter(event__end_time__gt=cutoff_time)
+                .first()
             )
 
             if next_event is None:
--- a/lego/apps/achievements/verification.py
+++ b/lego/apps/achievements/verification.py
@@ -35,13 +35,22 @@ def check_poll_responses(user: User, count: int):
 
 
 def check_longest_period_without_penalties(user: User, years: int) -> bool:
-    days = years * 365
-
-    if not (events := Registration.objects.filter(user=user).order_by("event__end_time")):
+    if not user.has_grade_group:
+        return False
+    if not (
+        events := Registration.objects.filter(
+            user=user, status=SUCCESS_REGISTER
+        ).order_by("event__end_time")
+    ):
         return False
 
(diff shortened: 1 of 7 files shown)


jonasdeluna:
This is very cool! Love the type changes.
But won't this check if the user has more than the given penalties thus incentivizing accumulating penalties?

aleksandervestlund:
Would you instead want to start at a level and go down as you receive more penalties? For example, start at gold level, but go down to silver after having received 3 penalties? That could also be a solution to remove the incentive. Initially, this was meant more as a progressive punishment, but that would be a creative approach.

jonasdeluna:
Hmm this is very much the opposite, but how about an achievement for going x time period without having any penalties (whilst still participating in some events)?
Like 1/2/3 years without penalties with 1 participation per semester?
Or less punshing time periods but something like that.
It could be more in line with the event rules.
I'll ask around on the meeting this wednesday in any case for any ideas.

aleksandervestlund:
I began writing a function to find the longest period without receiving a penalty. However, there is no logic to check if they attended an event in that time period. I think this could be quite easy to implement. Also, is `User.date_bumped` when a user left Abakus? In that case, it would make more sense to only find `timezone.now()` if this value is `None`.

aleksandervestlund:
I think how it is now would be good. It starts when a user has attended their first event and ends when they attended their last, to prevent people who have already left Abakus from getting the achievement. What do you think, @jonasdeluna?

jonasdeluna:
I think it's a lot better and would work well. Is there a way to check that the user additionally also attended at least one event every 365 days or something like that? In the case of a very inactive user that only attends something in their first semester first year and last semester last year they would still get it right?
If they attend one for every year at least I think it would be better.

jonasdeluna:
Oh and for the promotion to actually happen you need to make a function in promotions.py similar to the existing ones.

lego/apps/achievements/verification.py
```diff
@@ -32,6 +34,29 @@ def check_poll_responses(user: User, count: int):
     )
 
 
+def check_longest_period_without_penalties(user: User, years: int) -> bool:
+    days = years * 365
+
+    if not (events := Registration.objects.filter(user=user).order_by("end_time")):
```
jonasdeluna, line 40: `if not (events := Registration.objects.filter(user=user).order_by("event__end_time")):`

jonasdeluna (changes requested):

lego/apps/achievements/verification.py
aleksandervestlund: Is it the same for the next one, Penalty...?

jonasdeluna: No, all models that inherit from "BasisModel" will have created_at (and updated_at) attached to it.

jonasdeluna:
I tested this now, I think the filter should also check that the user is in the "Abakus" group, as alumni who are in the Orden would get it if not (tested on staging DB).

jonasdeluna:
In the current stats theres about 800 without a trophy, 500 with level 0, 300 with level 1, 100 with level 2 and 150 with level 3. Adding dual constraint of being a student (can check if has group 1-5 komtek/data) and event for each year (or something similar) would probably make the curve better.

aleksandervestlund:
> I tested this now, I think the filter should also check that the user is in the "Abakus" group, as alumni who are in the Orden would get it if not (tested on staging DB).

By using `user.is_abakus_member`?

aleksandervestlund:
> In the current stats theres about 800 without a trophy, 500 with level 0, 300 with level 1, 100 with level 2 and 150 with level 3. Adding dual constraint of being a student (can check if has group 1-5 komtek/data) and event for each year (or something similar) would probably make the curve better.

I wrote this function to check if they attend an event each year, but it returns `False` if they have not attended a single event for any of the active years. Do you think this is fair?

```python
for year in range(start_time.year, end_time.year + 1):
    if not any(
        event.event.end_time.year == year
        for event in events
        if event.status == SUCCESS_REGISTER
    ):
        return False
```

jonasdeluna:
Hmm I think ideally it should just not consider that time-period and move on to the next-best time period and check if that one matched the requirements. That way, you might get a smaller interval (than the one you were inactive in) rather than losing the trophy alltogether. So I guess it's like a longest common subsequence of years where you were registered and then pick the start and end to form the time-period.

aleksandervestlund:
I am not sure if my logic is entirely correct and have not tested the query. However, I believe it fixes the previous dilemmas.

jonasdeluna:
@aleksandervestlund I think this is good now. I checked in staging and it does run and is good. Please write some tests for this & run lint. I'll ask someone about how to run the CI runner for this (when its from a forked branch)
Regardless please speak to @falbru about getting the lego pin tomorrow for the christmas party as we think this is deserving of the pin even though it probably won't get merged before tomorrow. 
Really cool addition!

jonasdeluna:
Now that I think about it it might be too punishing with 1 registration per semester. How about changing it to 1 registration per year? Making it more possible for people who go on exchange as well.

aleksandervestlund:
> @aleksandervestlund I think this is good now. I checked in staging and it does run and is good. Please write some tests for this & run lint. I'll ask someone about how to run the CI runner for this (when its from a forked branch) Regardless please speak to @falbru about getting the lego pin tomorrow for the christmas party as we think this is deserving of the pin even though it probably won't get merged before tomorrow. Really cool addition!

Thanks❤️ Was a fun experience!

jonasdeluna:
I think to get the runner to run we'll have to first merge this into another branch and then make a PR to merge that into master.

jonasdeluna (approved):

jonasdeluna:
@aleksandervestlund The test appears to not pass, you can test that one specicially with python manage.py test lego.apps.achievements
Or did it pass for you?

jonasdeluna:
I think it fails on the 2nd one. And yeah the last one will probably fail too, as achievements are never revoked (and getting a penalty back in time like that shouldn't work):

```
❯ python manage.py test lego.apps.achievements
Found 16 test(s).
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
.......F........
======================================================================
FAIL: test_penalty_achievement (lego.apps.achievements.tests.tests.AchievementTestCase.test_penalty_achievement)
Test get penalty achievement.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/deluna/code/lego/lego/apps/achievements/tests/tests.py", line 411, in test_penalty_achievement
    self.assertTrue(
AssertionError: False is not true : User should have unlocked achievement for going 1 year without receiving a penalty, since there is less than a year between the events and it has not received any penalties.

----------------------------------------------------------------------
Ran 16 tests in 14.191s

FAILED (failures=1)
Destroying test database for alias 'default'...

```

jonasdeluna:
I tested it in local shell and it does give me the achievement so I believe the actual code works

image


image

jonasdeluna:
Ah you're probably right:

https://github.com/webkom/lego/blob/master/lego/apps/users/constants.py

FIRST_GRADE_DATA from lego.apps.users.constants

aleksandervestlund:
After closing the PR I don't have access to make changes anymore. I tried to change the bottom of the test to this:

```python
event = Event.objects.get(pk=2)
registration = Registration.objects.get_or_create(event=event, user=user)[0]
event.register(registration)
event.save()

penalty = Penalty.objects.create(
    user=user, reason="Ran beside the olympic swimming pool holding scissors."
)
penalty.created_at = now - timedelta(days=100)
penalty.save()

check_penalty_related_single_user(user)
self.assertFalse(
    Achievement.objects.filter(
        user=user, identifier=PENALTY_IDENTIFIER, level=0
    ).exists(),
    "User should not have unlocked achievement, since it has received "
    "a penalty in the year that it has attended events.",
)

penalty.created_at = now - timedelta(days=400)
penalty.save()

check_penalty_related_single_user(user)
self.assertTrue(
    Achievement.objects.filter(
        user=user, identifier=PENALTY_IDENTIFIER, level=0
    ).exists(),
    "User should have unlocked achievement for going 1 year without "
    "receiving a penalty, since there is less than a year between the "
    "events and it has not received any penalties.",
)
```

I also tried to refactor the code by replacing all instances of `"Abakus"` with `MEMBER_GROUP` and adding `FIRST_GRADE_KOMTEK` to the penalty test, both groups from `lego.apps.users.constants`.

jonasdeluna:
Hey I got the tests to work!

PR #3684, opened 2024-11-05
Implement meeting templates

jonasdeluna:
Resolves: ABA-846


jonasdeluna:
After review from Eik I will add all fields of meeting to the template so its back in WIP

lego/api/v1.py
```diff
@@ -151,6 +152,8 @@
     r"meeting-token", MeetingInvitationTokenViewSet, basename="meeting-token"
 )
 router.register(r"meetings", MeetingViewSet, basename="meeting")
+router.register(r"meetingtemplates", MeetingTemplateViewSet, basename="meetingtemplate")
```
ivarnakken, line 155:
```suggestion
router.register(r"meetings/templates", MeetingTemplateViewSet, basename="meetingtemplate")
```
I think /templates should be under `meetings. Also, meeting templates are two words

lego/apps/meetings/models.py
```diff
@@ -189,3 +189,29 @@ class ReportChangelog(BasisModel):
...
+    mazemap_poi = models.PositiveIntegerField(null=True)
+    report_author = models.ForeignKey(
+        User,
+        blank=True,
+        null=True,
+        related_name="meetingstemplates_reports",
+        on_delete=models.SET_NULL,
+    )
+    invited_users = models.ManyToManyField(User)
+    invited_groups = models.ManyToManyField(AbakusGroup)
```
ivarnakken, line 210: This is just copy paste from the Meeting model right? Is it possible to make this more scalable? Perhaps make a basis model both this and Meeting can inherit from? Perhaps even do `class MeetingTemplate(Meeting)` but you probably don't want everything from the meeting model 🤷🏼‍♂️

lego/apps/meetings/serializers.py
```diff
@@ -150,3 +155,34 @@ class Meta:
...
+            "location",
+            "start_time",
+            "end_time",
+            "description",
+            "mazemap_poi",
+            "report_author",
+            "invited_users",
+            "invited_groups",
+            "created_by",
+        )
```
ivarnakken, line 188: Again, this is basically copy paste from a meeting serializer, right?

ivarnakken (changes requested):
I love that you wrote tests! 😍

lego/apps/meetings/models.py
jonasdeluna: They are slightly different because the meeting needs additional fields on the invited_users many-to-many

lego/apps/meetings/serializers.py
jonasdeluna: Only report and report_author are the same. I could move it out to a different basismeetingserializer class or something but I dont see that much point?

PR #3686, opened 2024-11-10
Add leaderboard and overview of achievements

jonasdeluna:


eikhr (approved):

eikhr (commented):
I think this might become slow with all users in prod, but I'm not sure how to improve it...

lego/apps/achievements/utils/calculation_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    EVENT_PRICE_IDENTIFIER: [2, 3, 5],
+    MEETING_IDENTIFIER: [1],
+    POLL_IDENTIFIER: [0, 2, 4],
+    PENALTY_IDENTIFIER: [0, 3, 5, 6],
+}
+
+delta = 0.1
+# Remember to update this rarity list when adding new achievement
+MAX_POSSIBLE_SCORE = sum(
+    max(rarity_list) + 1 + (max(rarity_list) * delta)
```
danielyanghansen, line 24: Why does this work? `rarity_list` doesn't seem to be defined at this point

danielyanghansen: I think this is going to give you a runtime error. If this code works, it seems to be because of unintended behavior

lego/apps/achievements/utils/calculation_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    score = 0.0
+
+    user_achievements = user.achievements.all()
+    for achievement in user_achievements:
+        rarity_list = ACHIEVEMENT_RARITIES.get(achievement.identifier, [])
+
+        value = rarity_list[achievement.level]
+        score += value + 1 + (achievement.level * delta)
+
+    return score if MAX_POSSIBLE_SCORE else 0
```
danielyanghansen, line 40: I'm not quite sure why the presence of `MAX_POSSIBLE_SCORE` would matter here

lego/apps/users/models.py
```diff
@@ -421,6 +422,12 @@ def delete(self, using=None, force=False):
                     event.unregister(event.registrations.get(user=self))
         super(User, self).delete(using=using, force=force)
 
+    def save(self, *args, **kwargs):
+        from lego.apps.achievements.utils.calculation_utils import calculate_user_rank
```
danielyanghansen, line 426: Could you educate me on the benefits of a late import like this? :angel:

danielyanghansen (changes requested):
Overall looks good.
Pagination of the `LeaderBoardViewSet.get_queryset` would be nice, but can be done in a later PR.

lego/apps/achievements/utils/calculation_utils.py
jonasdeluna: This is apparently due to a quirk in python as the defenition depends on a dictionary that already exists it apparently works

PR #3689, opened 2024-11-13
Edit student contacts

falbru:
Related PR: https://github.com/webkom/lego-webapp/pull/5053


lego/apps/companies/serializers.py
```diff
@@ -36,9 +39,12 @@ def create(self, validated_data):
 
 
 class StudentCompanyContactSerializer(BasisModelSerializer):
+    user = PublicUserField(queryset=User.objects.all())
+    semester = SemesterField(queryset=Semester.objects.all())
```
ivarnakken, line 43: Are these querysets needed? If so, why

ivarnakken (approved):
Nice that you added tests!

lego/apps/companies/serializers.py
falbru: no you're right they're not needed

falbru: or after checking: for fields, you actually do need an explicity query set. Or else it causes an error

ivarnakken: hmm okay not sure if I get why - you'd think it would be redundant

falbru: ik but that's how it's used throughout the whole code base

PR #3697, opened 2024-12-11
Include user reg to events in event list

ivarnakken:
Needed for ABA-1196


eikhr (approved):
Looks good, you should probably check if we are using prefetching for this query and add this if so:)

ivarnakken:
> Looks good, you should probably check if we are using prefetching for this query and add this if so:)

Already implemented! Excellent code 😎 
https://github.com/webkom/lego/blob/master/lego/apps/events/views.py#L129-L135

PR #3702, opened 2024-12-15
Use pre-defined choices for company types

ivarnakken:
Can be used in filtering on /companies, and to auto-complete the interest form after a company has been selected (the company interest model already uses the same choices).

The text field is not optional in the frontend, so it has not been used that much. I will manually migrate these.


norbye (approved):
Nice!

eikhr (approved):

falbru (approved):
LGTM

PR #3703, opened 2024-12-15
Fix external sync for google

norbye:
Debugging this was pretty annoying, as stuff that worked locally did not work in prod because we apparently don't use the same versions for dependencies in prod that we do locally.. This is now fixed.

While the locking would've solved the problem for now - the package we used to use `oauth2client` has been deprecated since 2019 heheh so I got rid of that and implemented its successor.

Hard to perform exact tests on prod-only behaviour, but I've run the snippets of code I've changed in prod and it worked, so it should be good, although I'll have to do a deployment after merging to be completely sure.

Resolves ABA-1202


ivarnakken (approved):
Makes sense! Good job! 🤩

PR #3708, opened 2024-12-25
Add pagination, ordering to BDB

Bestem0r:
# Description

Adds pagination and ordering by query parameters to BDB Pro. Missing ordering options and broken status filtering will be readded and fixed in a different PR.  

[Related front end PR ](https://github.com/webkom/lego-webapp/pull/5254)

# Result

- [x] Changes look good on both light and dark theme.
- [x] Changes look good with different viewports (mobile, tablet, etc.).
- [x] Changes look good with slower Internet connections.

https://github.com/user-attachments/assets/e4639b09-b3e9-4594-9a9b-a3d1c6997b7d

# Testing

Navigate to BDB Pro and ensure that content is paginated correctly as you scroll down.

Resolves [ABA-1138](https://linear.app/abakus-webkom/issue/ABA-1138/add-pagination-to-bdb)


lego/apps/companies/views.py
```diff
@@ -46,8 +46,10 @@
 
 class AdminCompanyViewSet(AllowedPermissionsMixin, viewsets.ModelViewSet):
     queryset = Company.objects.all().prefetch_related("semester_statuses", "files")
-    pagination_class = None
     permission_handler = CompanyAdminPermissionHandler()
+    filter_backends = [filters.OrderingFilter]
+    ordering_fields = ["name", "created_at"]
+    ordering = "name"
```
ivarnakken, line 52: Can we move this to `lego/apps/companies/filters.py`, like we do for the other view sets

ivarnakken (changes requested):
Nice!

lego/apps/companies/filters.py
```diff
@@ -17,3 +18,8 @@ class CompanyInterestFilterSet(FilterSet):
     class Meta:
         model = CompanyInterest
         fields = ("semesters", "events")
+
+
+class CompanyOrderingFilter(filters.OrderingFilter):
```
ivarnakken:
```suggestion
class AdminCompanyFilterSet(FilterSet):
```

lego/apps/companies/views.py
```diff
@@ -47,9 +51,7 @@
 class AdminCompanyViewSet(AllowedPermissionsMixin, viewsets.ModelViewSet):
     queryset = Company.objects.all().prefetch_related("semester_statuses", "files")
     permission_handler = CompanyAdminPermissionHandler()
-    filter_backends = [filters.OrderingFilter]
-    ordering_fields = ["name", "created_at"]
-    ordering = "name"
+    filter_backends = [CompanyOrderingFilter]
```
ivarnakken:
```suggestion
    filterset_class = AdminCompanyFilterSet
```

Not sure if I get the difference, but nice to be consistent and use the `django_filters` library

ivarnakken (changes requested):

lego/apps/companies/filters.py
Bestem0r: Hmm I can't seem to get ordering to work with a `FilterSet` and I can't find any examples of it done in the codebase either. If you look at e.g. `events/views.py`, ordering is done using `rest_framework` `filters`, while normal filtering is done using `django_filters` `FilterSet`. Since the `AdminCompanyViewSet` does not have any normal filtering, wouldn't it make sense to revert it to not have the filters `companies/filters.py`?

ivarnakken: ah yes I was a bit quick here - I guess it makes sense to use the `rest_framework` `filters` since we're really only doing ordering here - sorry!

Bestem0r:
Reverted to how it was previously now @ivarnakken

lego/apps/companies/models.py
```diff
@@ -57,6 +57,7 @@ class Company(BasisModel):
 
     class Meta:
         permission_handler = CompanyPermissionHandler()
+        ordering = ["name"]
```
ivarnakken, line 60: Why is this needed?

Bestem0r: It's the default value. Without it, the front end crashes when accessing bdb without providing ordering in the query params

ivarnakken (approved):
Nice!

PR #3715, opened 2025-01-22
Improve feedback for abacard responses

norbye:
Apparently the `error_code` (which is what we use in lego-webapp to write the norwegian output) returned `already_present` if the user was already present, not properly registered, or waitlisted, which not surprisingly made it confusing for admins scanning IDs.

Split it into;
* unregistered user
* not properly registered user
* waitlisted user
* user already registered as late or absent
* user already present

Only new check should be the payment check

---

All possible `error_code` values are now;
* no_user
* not_registered
* unregistered
* not_properly_registered
* waitlisted
* already_present
* late_or_absent
* missing_payment

Part 1/2 of resolving ABA-1233 (missing lego-webapp PR thus far)

TODO
- [x] Write some tests
- [ ] Wait for @falbru to publish the frontend part of it ❤️


ivarnakken (approved):
Nicely tested!

PR #3716, opened 2025-01-22
Remove start-up event and revue collab from constants

viljeno:
Updated interest form as requested by Intercom.

ABA-1232


ivarnakken:
We probably want to use these fields later, so we probably shouldn't remove this from the backend model. This would actually remove these values from previous interest forms. Code looks good, but I suggest closing this PR since it's not something we should remove.

ivarnakken:
Just to clarify: updating the frontend was enough! 😄

PR #3719, opened 2025-01-24
Prevent admins from searching meetings

norbye:
It's been bugging a little that admins can search all meetings (even though we can't actually see the meetings - which is an annoying bug/feature caused by the view filtering the queryset instead of using the permissions)

Anyways, this change ensures that no users have permissions to meetings that they are not a part of, resulting in them being excluded from searches. It might seem like searching was the only place where meetings was accessed not through the meetings view, so that's why that is the only place it was noticeable.

Works just fine locally (the creator can still view and edit the meeting, and the superadmin cannot view it, nor search for it)

### DIFF
--- a/lego/apps/meetings/permissions.py
+++ b/lego/apps/meetings/permissions.py
@@ -28,8 +28,15 @@ class MeetingPermissionHandler(PermissionHandler):
         if not user.is_authenticated:
             return False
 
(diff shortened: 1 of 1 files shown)


ivarnakken (approved):
Finally!

PR #3720, opened 2025-01-24
Revert crypt hash generation on login

norbye:
We've been generating crypt hashes on login for users missing it due to it not being set on registration until 8 months ago. Now we're down from 487 users with no hash to 78, so I figured it could be time to remove the temporary update.

If any of the remaining 78 users against all odds require the hash they can reset their password.

Originally implemented in https://github.com/webkom/lego/pull/3602 and https://github.com/webkom/lego/pull/3603

Resolves ABA-942

### DIFF
--- a/lego/api/tests/test_authentication.py
+++ b/lego/api/tests/test_authentication.py
@@ -38,32 +38,6 @@ class JSONWebTokenTestCase(BaseAPITestCase):
         self.assertContains(response, text="token", status_code=status.HTTP_201_CREATED)
         self.assertContains(response, text="user", status_code=status.HTTP_201_CREATED)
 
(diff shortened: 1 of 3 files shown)


ivarnakken (approved):
Any reason to not just wait a little with merging?

norbye:
> Any reason to not just wait a little with merging?

Not really, just wanted to make sure it wasn't eventually forgotten. And I assumed that there wouldn't be that many users that haven't logged in during the last semester that will login this semester, and if I delay this until august I won't be here to remember it 🤷

Arashfa0301 (approved):
Good to go

jonasdeluna:
wow someone else is also up late

PR #3726, opened 2025-02-05
Update setup docs to work with new poetry versions

eikhr:
Poetry expects the venv folder to be named ".venv" for some reason...


norbye (approved):
👍

ivarnakken (approved):

PR #3728, opened 2025-02-05
Prevent non-abakus members from joining interest groups

sondrejk:
Solves [ABA-625](https://linear.app/abakus-webkom/issue/ABA-625/)

### DIFF
--- a/lego/apps/users/views/memberships.py
+++ b/lego/apps/users/views/memberships.py
@@ -1,8 +1,10 @@
 from django_filters.rest_framework import DjangoFilterBackend
-from rest_framework import filters, viewsets
+from rest_framework import filters, status, viewsets
+from rest_framework.response import Response
 
 from lego.apps.permissions.api.filters import LegoPermissionFilter
 from lego.apps.permissions.api.views import AllowedPermissionsMixin
+from lego.apps.users import constants
 from lego.apps.users.filters import MembershipFilterSet
 from lego.apps.users.models import AbakusGroup, Membership
 from lego.apps.users.serializers.memberships import MembershipSerializer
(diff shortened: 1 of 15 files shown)


sondrejk:
The first force-pushes were to add co-author. Took some time to figure out...

lego/apps/users/views/memberships.py
```diff
@@ -32,4 +34,12 @@ def get_queryset(self):
...
     def create(self, request, *args, **kwargs):
         request.data["abakus_group"] = kwargs["group_pk"]
+        user_groups = [
+            membership.abakus_group.name for membership in request.user.memberships
+        ]
+        group_id = kwargs["group_pk"]
+        group = AbakusGroup.objects.get(pk=group_id)
+        if group.type == constants.GROUP_INTEREST and "Abakus" not in user_groups:
+            return Response(status=status.HTTP_403_FORBIDDEN)
+
```
norbye, line 44: This could be simplified slightly by using the `is_abakus_member` property:) (as defined in https://github.com/webkom/lego/blob/master/lego/apps/users/models.py#L213-L215)

```suggestion
        group = AbakusGroup.objects.get(pk=kwargs["group_pk"])
        if group.type == constants.GROUP_INTEREST and not request.user.is_abakus_member:
            return Response(status=status.HTTP_403_FORBIDDEN)

```

norbye (changes requested):
Awesome stuff!

What you've done looks like it would work beautifully, and we already have a property that does pretty much exactly this:) also it's adviseable to always try to use constants instead of strings where possible (constants.MEMBER_GROUP instead of "Abakus" - like how it's done in the `is_abakus_member` property)

Would also be awesome with a few tests to ensure this works as expected ❤️

lego/apps/users/views/memberships.py
tindevw: I tried using the is_abakus_member property as suggested but non-students were still able to join. Switching to has_grade_group resolved the issue, though I'm unsure if that was the intended approach.

ivarnakken (approved):
Looks solid! 😍 Awesome job

Would be nice with some simple tests here to test this

lego/apps/users/views/memberships.py
norbye: Ahh I see, I suggested you use `is_abakus_member` because that was what your code checked previously. However, a user can be an abakus member without being a student, so if what you wanted to accomplish was to only allow students to register to groups then you're right, `has_grade_group` would be a better approach!

norbye (approved):
Sweet!

Seems like you need to do some quick formatting, and some tests would be nice but other than that it looks good

ch0rizo (approved):
Nice!

PR #3727, opened 2025-02-05
Remove comment field from companies

Bestem0r:
New attempt since the last PR went stale (and broken).

PR #3730, opened 2025-02-09
Remove admin_comment from fixtures

norbye:
These fixtures wouldn't work after https://github.com/webkom/lego/pull/3727 was merged, but now they do


ivarnakken (approved):

Bestem0r (approved):
Good catch!

PR #3731, opened 2025-02-09
Add most important groups from prod to fixtures

norbye:
About time that more of our core groups are added to development fixtures


ivarnakken (approved):
I like!

PR #3743, opened 2025-02-15
Patch logic so it doesnt crash

jonasdeluna:
# Description

Fixes bug where the promotion would crash if someone moved up or down a rank due to having the trophy prior.
Also fixes proper counting of events for this trophy.

# Testing

Resolves ... (either GitHub issue or Linear task)


norbye (approved):
There's been a few patch PRs in a row now - could it maybe be an idea to write some more tests for some more scenarios to avoid this? 👉🏻👈🏻🥺

jonasdeluna:
Good idea, I will make some tests before merging this 👍👍

PR #3744, opened 2025-02-17
refactor: shorten registration events

### LINKED ISSUE #3714: Registration events collide with lectures
## The problem

Currently, it is possible to subscribe to a calendar on the [website](https://abakus.no) which adds the registration times for all events to your personal calendar. 

All these registration events are 30 minutes long. For an event whose registration opens at 12:00, the event will be added in the calendar from 12:00 to 12:30. 

The problem with this is that most events at NTNU start at a quarter past. For example, lectures can start at 12:15. Because of this, the registration event will push the class out of place in a calendar like this:

problem_example

For me, this is pretty annoying and makes my calendar noisy, so I tend to unsubscribe from registration events. But then that makes me miss the next registrations. As a result, I have to re-enable these events only when I'm waiting on one.

## Suggested solution

Reduce registration events in the calendar from 30 to 15 minutes. This will prevent them from colliding with other NTNU events in the calendar. This way, both events will fit linearly into the calendar and avoid the weirdness of the previous picture. Here's an example:

possible_solution

jovialen:
# Shorten registration events

Resolves #3714. Shortens registration events from 30 to 15 minutes to better fit into NTNU calendar.


jonasdeluna:
Awesome that you want to contribute 👍👍
We will probably discuss this first before potentially merging

jovialen:
It's just an annoyance of mine. Personally, I'd prefer the registration events to look a bit thin over obstructing my classes in the calendar. I'll cross my fingers that you'll agree :)

eikhr:
We discussed this and did some testing in our calendars. It seems like the event could be even shorter than 15 mins and still be displayed fine in our calendars. Is there any reason why you went with 15, instead of 5 (or even 0!) minutes?

jovialen:
I choose 15 minutes simply because it was the smallest reduction which would fit the registration events into the between-lectures slot in the calendar. Further reducing it to 0 minutes would probably be completely fine, but I chose not to do so for two reasons:

1. I wasn't sure it wouldn't cause problems with some calendars
2. Doing so would also warrant refactoring the code to remove the registration completely, and I wanted to avoid making big changes to the codebase uninvited.

If these reasons aren't concerning, I could probably add the aforementioned refactor to this pull request.

jovialen:
Is this pull request still being considered?

ch0rizo (approved):

PR #3767, opened 2025-03-05
Add banner to backend

jonasdeluna:
# Description

Adds banners

# Testing

Resolves ABA-1328


lego/apps/banners/permissions.py
```diff
@@ -0,0 +1,24 @@
...
+        perm,
+        obj=None,
+        queryset=None,
+        check_keyword_permissions=True,
+        **kwargs
+    ):
+        if perm == LIST:
+            return True
+        if perm == VIEW:
+            return True
```
itsisak, line 20:
```suggestion
        if perm in [LIST, VIEW]:
            return True
```

lego/apps/banners/serializers.py
```diff
@@ -0,0 +1,29 @@
...
+            Banners.objects.exclude(pk=instance.pk).update(current_private=False)
+        if validated_data.get("current_public", False):
+            Banners.objects.exclude(pk=instance.pk).update(current_public=False)
+        return super().update(instance, validated_data)
+
+    def create(self, validated_data):
+        if validated_data.get("current_private", False):
+            Banners.objects.all().update(current_private=False)
+        if validated_data.get("current_public", False):
+            Banners.objects.all().update(current_private=False)
```
itsisak, line 28: Maybe forgot something here? Both updates do the same..

lego/apps/banners/models.py
```diff
@@ -0,0 +1,19 @@
...
+    link = models.CharField()
+    current_public = models.BooleanField(default=False, null=False, blank=False)
+    current_private = models.BooleanField(default=False, null=False, blank=False)
+
+    class Meta:
+        permission_handler = BannersPermissionHandler()
+        indexes = [
+            models.Index(fields=["current_public"]),
+            models.Index(fields=["current_private"]),
+        ]
```
itsisak, line 19: Just a thought, but I think it might be possible to add `meta.constraints` to ensure only ever one instance ever can have `current_public` set to true. It is strictly not needed if it is always handled correct, but could maybe be nice to enforce it on a database-level

lego/apps/banners/serializers.py
```diff
@@ -0,0 +1,29 @@
...
+            "current_private",
+            "current_public",
+        )
+
+    def update(self, instance, validated_data):
+        if validated_data.get("current_private", False):
+            Banners.objects.exclude(pk=instance.pk).update(current_private=False)
+        if validated_data.get("current_public", False):
+            Banners.objects.exclude(pk=instance.pk).update(current_public=False)
+        return super().update(instance, validated_data)
```
eikhr, line 22: Shouldn't these check if validated_data `current_private` and `current_public` are **True**? Or am I misunderstanding what this does?

Also write some simple tests for the actual logic parts at least!

lego/apps/banners/views.py
```diff
@@ -0,0 +1,50 @@
...
+    lookup_value_regex = r"\d+"
+    serializer_class = BannersSerializer
+    queryset = Banners.objects.all()
+    permission_classes = [LegoPermissions, AllowAny]
+
+    def get_queryset(self):
+        if self.action == "current_private":
+            return Banners.objects.filter(current_private=True)
+        if self.action == "current_public":
+            return Banners.objects.filter(current_public=True)
```
eikhr, line 21: Why do we do this here, instead of directly filtering in the action handlers? And does it make a difference using `Banners.objects.filter...` instead of `super().get_queryset().filter...`?

lego/apps/banners/views.py
```diff
@@ -0,0 +1,50 @@
...
+        detail=False,
+        methods=["get"],
+        url_path="current_public",
+        permission_classes=[AllowAny],
+    )
+    def current_public(self, request, *args, **kwargs):
+        queryset = self.get_queryset()
+        banner = queryset.first()
+        if banner is None:
+            return Response({"detail": "Not found."}, status=status.HTTP_404_NOT_FOUND)
```
eikhr, line 48: Do we want this to return a 404 error? Is it possible to instead just do a 200 ok with no data? I think that would make more sense, as it most likely just means there is no active banner at this time, not that there was an error.

eikhr (commented):
This has been talked about for soo long, looking forward to finally getting it implemented!

lego/apps/banners/permissions.py
```diff
@@ -0,0 +1,21 @@
+from lego.apps.permissions.constants import LIST, VIEW
+from lego.apps.permissions.permissions import PermissionHandler
+
+
+class BannersPermissionHandler(PermissionHandler):
+    default_keyword_permission = "/sudo/admin/banners/{perm}/"
```
itsisak, line 6:
```suggestion
```

I think this is what the `keyword_permission` is by default anyways.

lego/apps/banners/permissions.py
```diff
@@ -0,0 +1,21 @@
...
+        perm,
+        obj=None,
+        queryset=None,
+        check_keyword_permissions=True,
+        **kwargs
+    ):
+
+        return super().has_perm(
+            user, perm, obj, queryset, check_keyword_permissions, **kwargs
+        )
```
itsisak, line 21:
```suggestion
```

This seems redundant

lego/apps/banners/tests.py
```diff
@@ -0,0 +1,133 @@
...
+            "header": "Test Banner",
+            "subheader": "Test Subheader",
+            "link": "http://example.com",
+            "color": "red",  # assuming "red" is a valid option
+            "current_public": False,
+            "current_private": False,
+        }
+
+    def test_banner_create_without_permission(self):
+        """
```
itsisak, line 48: Maybe ensure the other default endpoints also require permission since it seems `current_private` and `current_public` are the only ones that should ever be available for non-admins?

itsisak (approved):
Looks good to me! Will be nice to have this feature!

PR #3773, opened 2025-03-07
Add "sub" field to OAuth2 userdata

### LINKED ISSUE #3724: Misconfigured OAuth spec
## Problem

It seems the LEGO's implementation of OAuth tries to read `scopes` instead of `scope` when deserializing request.

It's just a hunch and not verified. 

## Scope of issue

It means I cannot add Abakus as an OAuth source in Authentik to facilitate *log in with Abakus* for internal services for Teknikkgruppa-revy.

## Resources

[RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749#appendix-A.4)

eikhr:
OIDC uses "sub" as the id-field. While this is still not really OIDC complient (because we don't use the "openid" scope) it might fix issues with some OAuth2 clients

# Testing
- [ ] I have thoroughly tested my changes.

Resolves #3724 (maybe)

### DIFF
--- a/lego/apps/users/serializers/users.py
+++ b/lego/apps/users/serializers/users.py
@@ -142,12 +142,16 @@ class Oauth2UserDataSerializer(serializers.ModelSerializer):
     is_student = serializers.SerializerMethodField()
     is_abakus_member = serializers.BooleanField()
 
+    def get_sub(self):
+        return str(self.instance.id)
+
     def get_email(self):
         return self.instance.email_address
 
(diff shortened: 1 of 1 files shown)


ivarnakken (approved):

jonasdeluna (approved):

jonasdeluna (commented):
Do you not need to add it as a serializer method field?
but I guess the tests pass so its ok?

PR #3776, opened 2025-03-07
Fix conflicting survey migrations and make banner subheader optional

itsisak:
# Description

There was multiple migrations on survey app with same number, this lead do conflicting migrations. Important fix!!!

Also make subheader on banner optional


eikhr (approved):
Looks good!

PR #3777, opened 2025-03-08
Create the feature flags

jonasdeluna:
# Description

Implement the feature flags

# Testing

Resolves ABA-1345


ShaileshS1702 (approved):
LGTM!

lego/apps/featureflags/tests.py
norbye, line 1: Could you write some tests for what happens to users that are not part of the group that should be able to see the feature flag, but have `/sudo/admin/` permissions?

For this sort of feature I imagine that we would prefer granular access that allows us to see it as we know what it does - but we might not want Ledelsen to see it as they might not know what has been feature flagged and not, which could cause confusion

jonasdeluna: Ah, wait so you want us to see it but not ledelsen? The current behaviour does not take into account any permissions when checking the featureflag so unless Webkom or Ledelsen is specified in the groups, they will also get false. Did you want to change that behaviour or only a test of the current behaviour?

jonasdeluna: I also realized we do not have /sudo/ permission (anymore?) we also have /sudo/admin

norbye: Ah nice, I just assumed you interfered with the permission system somehow - then it's probably good

PR #3778, opened 2025-03-09
Make email lowercase when saving model and validating

jonasdeluna:
# Description

Makes all emails lowercase when saving.
All current emails have been set to lower using shell.
There was one guy with duplicate email but he was deleted.

# Testing

Resolves ABA-374

### DIFF
--- a/lego/apps/users/models.py
+++ b/lego/apps/users/models.py
@@ -420,7 +420,9 @@ class User(
 
     def save(self, *args, **kwargs):
         self.email = self.email.lower()
-        self.student_username = self.student_username.lower() if self.student_username else None
+        self.student_username = (
+            self.student_username.lower() if self.student_username else None
+        )
         if self.pk:
             from lego.apps.achievements.utils.calculation_utils import (
                 calculate_user_rank,
--- a/lego/apps/users/models.py
+++ b/lego/apps/users/models.py
@@ -391,10 +391,6 @@ class User(
     class Meta:
         permission_handler = UserPermissionHandler()
 
-    def clean(self):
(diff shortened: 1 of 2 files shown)


lego/apps/users/models.py
```diff
@@ -392,6 +392,7 @@ class Meta:
         permission_handler = UserPermissionHandler()
 
     def clean(self):
+        self.email = self.email.lower()
```
norbye, line 395: Any reason why you added this to both `clean()` and `save()` when `self.student_username` only appears to be done in `clean()`? Would think that `clean()` would be sufficient, and if it's not, that `self.student_username` should also be handled in `save()`?

Or are there any other differences that I'm not aware of?

jonasdeluna: Actually I'm not sure if we actually use clean anywhere in the code when taking a second look?
Reading up clean is only called for form validations which we dont use?
https://docs.djangoproject.com/en/5.1/ref/forms/validation/
So perhaps it should only be in save().
I added it just to make sure though.

norbye: In that case I would suggest you delete the clean function and add the student_username thingy to save so we lowercase that as well. And maybe check wherever we use the student_username to verify uniqueness that it case sensitive. (And maybe check if there are any uppercase student usernames in the db to check if its really been working)

Would also be nice to remove other uses of clean() if it does nothing

ivarnakken (approved):

norbye (approved):

jonasdeluna:
There are currently no users with student_username uppercase so this does not seem to be an issue. will merge this now.

PR #3788, opened 2025-03-17
Allow non admins to list surveys and add user filter

itsisak:
# Description

Allow users with `/sudo/admin/surveys/list/` to list surveys. (This permission should be added to `Abakus` group. This way all users should be able to list surveys. Filtering ensures they still only can list surveys they should have access to. 

# Testing

Resolves ABA-1358

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced survey filtering that allows users to refine survey views based on attendance.
  - Updated survey access so that authenticated users—including regular and attended users—can now view their own surveys, while admin users retain full access.
  - Added a new permission path for the `Abakus` group, enabling access to a new endpoint for listing surveys.

These improvements streamline the survey experience while expanding access for authorized users.

### DIFF
--- a/lego/apps/surveys/filters.py
+++ b/lego/apps/surveys/filters.py
@@ -1,14 +1,23 @@
-from django_filters import CharFilter, FilterSet
+from django.db.models import Q
+from django_filters import CharFilter, FilterSet, NumberFilter
 
+from lego.apps.events.constants import PRESENCE_CHOICES
 from lego.apps.surveys.models import Submission, Survey
 
 
 class SurveyFilterSet(FilterSet):
     company = CharFilter("event__company")
+    user = NumberFilter(method="filter_by_attended_user")
 
     class Meta:
         model = Survey
         fields = ("event",)
+
+    def filter_by_attended_user(self, queryset, name, value):
+        return queryset.filter(
+            Q(event__registrations__user=value)
+            & Q(event__registrations__presence=PRESENCE_CHOICES.PRESENT)
+        )
 
 
 class SubmissionFilterSet(FilterSet):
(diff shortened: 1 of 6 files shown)


lego/apps/surveys/filters.py
```diff
@@ -1,15 +1,24 @@
-from django_filters import CharFilter, FilterSet
+from django.db.models import Q
+from django_filters import CharFilter, FilterSet, NumberFilter
 
+from lego.apps.events.constants import PRESENCE_CHOICES
 from lego.apps.surveys.models import Submission, Survey
 
 
 class SurveyFilterSet(FilterSet):
     company = CharFilter("event__company")
```
jonasdeluna, line 9: Does this work?

jonasdeluna (approved):
Nice!
Should there be some filtering so that the organizers can only see surveys they have edit access to for the event instead of seeing all of them if you are an admin?

lego/apps/surveys/filters.py
itsisak: Do you mean the filtering? If so yes, seems to work fine, just add `user=<user_id>` and you will only get surveys that user attended

jonasdeluna: Ah i meant the company one actually but maybe it wasnt changed here

itsisak:
> Nice! Should there be some filtering so that the organizers can only see surveys they have edit access to for the event instead of seeing all of them if you are an admin?

Hmm, yes this would probably be nice, but thus far I have tried to keep most of the previous functionality how it is

lego/apps/surveys/filters.py
itsisak: I see, that was there before, don't know what it is used for and haven't tested it

PR #3789, opened 2025-03-17
Add gala achievement

aleksandervestlund:
* Primarily developed by @karolipb.
* Add gala achievements to encourage attending them.
* The numbers for the different levels are very loose.
* At this moment in time, all galas count, but it could also just be "studentgallaer".
* Have no idea if it works.
* See frontend PR webkom/lego-webapp#5447.


jonasdeluna:
I like the initiative but I have some concerns. 
I feel like galas tend to be more abakom-focused, so it might just turn out to be a "who is/has been a committee member (the longest)" achievement?
The halvingfest was very anti-galla imo, I mean everyone wore galla outfits but it was mostly a mess. Maybe just my experience tho.
Recently we (I) wrote some (ironcally) internal notes on the achievement system but I can share the markdown files here:

[New.md](https://github.com/user-attachments/files/19298229/New.md)
[Guidelines.md](https://github.com/user-attachments/files/19298230/Guidelines.md)

lego/apps/achievements/verification.py
```diff
@@ -137,6 +137,29 @@ def check_total_genfors_events(user: User, count: int) -> bool:
...
+    # Will need to be updated if there are more gala-like events
+    gala_substrings = (
+        "bankett",  # Max 3 (itDAGENE)
+        "halvingfest",  # Max 1
+        "immatrikuleringsball",  # Max 5
+        "jubileum",  # Max 4 (Abakus, Abakusrevyen, LaBamba, readme)
+        "julebord",  # Max 5
+        "utmatrikuleringsfest",  # Max 1
+        "vaargalla",  # Max 5
+    )
```
jonasdeluna, line 150: Move this to constants.py?

lego/apps/achievements/verification.py
```diff
@@ -137,6 +137,29 @@ def check_total_genfors_events(user: User, count: int) -> bool:
...
+
+    query = Q()
+    for substring in gala_substrings:
+        query |= Q(event__title__icontains=substring)
+
+    queryset = _passed_user_registrations(user).filter(query)
+
+    if count == 1:
+        return queryset.exists()
```
jonasdeluna, line 159: I feel like this can be removed as it's just the same as below?

jonasdeluna:
Also as described in the guide, please add it to raritylist :)

lego/apps/achievements/verification.py
aleksandervestlund: That was just meant as a small optimization so that if the queryset is very large, it doesn't need to find the exact length, but I get your point🦥

aleksandervestlund:
You make some valid points, but I feel that many members not in Abakom attend the different galas (maybe except for Vaargalla)—especially the itDAGENE banquet, Julebord and the different jubilees. Many first graders also have Immatrikuleringsfest as their first event, so this may motivate them further to attend. I kind of agree that Halvingfest and Utmatrikuleringsfest are kind of borderline galas, but in the description the dress code is described as gala, so I wasn't sure🤷🤷

ivarnakken:
> I feel like galas tend to be more abakom-focused, so it might just turn out to be a "who is/has been a committee member (the longest)" achievement?

Imo I don't see galla as something that is only for Abakom, quite the opposite.

I agree that halvingfest should not count

ollfkaih:
Agree with attending galas as a potentially nice incentive to more non-Abakom to attend them.

> Good job! But please call it "galla" with two Ls 😆
Unless we are looking to start using Norwegian variable names, which we only currently do for specific and difficult translations (e.g. genfors), I disagree. [[1]](https://en.wiktionary.org/wiki/gala) [[2]](https://www.merriam-webster.com/dictionary/galas) [[3]](https://dictionary.cambridge.org/us/dictionary/learner-english/gala)

aleksandervestlund:
To fix the spelling is really simple, but what's the conclusion? We can probably delete "Halvingfest" and "Utmatrikuleringsfest" either way since they're not "studentergalla"s

ivarnakken:
> Unless we are looking to start using Norwegian variable names, which we only currently do for specific and difficult translations (e.g. genfors), I disagree. [[1]](https://en.wiktionary.org/wiki/gala) [[2]](https://www.merriam-webster.com/dictionary/galas) [[3]](https://dictionary.cambridge.org/us/dictionary/learner-english/gala)

... okay fair enough 😓

jonasdeluna:
> To fix the spelling is really simple, but what's the conclusion? We can probably delete "Halvingfest" and "Utmatrikuleringsfest" either way since they're not "studentergalla"s

The meeting is 16:15 today we may be able to come to a conclusion then 👍👍

jonasdeluna:
After a long discussion we have decided to merge this.

Karolipb:
Glad to hear!! 🤗
With or without "halvingfest" and "utmatrikuleringsfest"? 
And the rarities in the front end are probably incorrect and we haven't run `makemigrations`.

jonasdeluna:
Probably without halvingfest and utmatrikuleringsfest. The rarities will probably be reworked soon anyways to have less rarities.

jonasdeluna:
@Karolipb If you want the lego pin you should probably message @magnusbrecke on slack or something I think he is responsible for that as he's the leader :+1:

PR #3791, opened 2025-03-18
Fix comment notifications

eikhr:
# Description

Fixes comment notifications by removing a broken permission check. I don't believe it is necessary, but please tell me if you know what that did!

I don't like emails on new comments/comment replies, but imo. it doesn't hurt having it as an option. I do however feel like we should shell and disable it for all current users, as well as making it default disabled.

# Testing

Resolves ABA-1395


itsisak (approved):

eikhr (commented):
<s>Email disabled by default doesn't work...</s> Edit: No longer needed, as I removed new comment notifications

ivarnakken (approved):

PR #3809, opened 2025-03-26
Implement email generation for lendingRequest objects

Arashfa0301:
# Description

Implemented email generation for lendingRequest objects. Every user that is in the responsible users list for a lendingRequest objects receives the mail


lego/apps/lending/serializers.py
```diff
@@ -211,3 +213,13 @@ def update(self, instance, validated_data):
             )
 
         return instance
+
+    def create(self, validated_data: CreateLendingRequestType) -> LendingRequest:
+        instance: LendingRequest = super().create(validated_data)
+
+        for lender in instance.lendable_object.can_edit_users.all():
```
jonasdeluna, line 220: Maybe there should be some check to send the email to edit groups too?

jonasdeluna (approved):
Awesome! :chart_with_upwards_trend:

PR #3810, opened 2025-03-26
Add the new timeline entry object

jonasdeluna:
# Description

Replaces comments on lending requests with timeline entries.
They have validation so you cannot create new timeline objects for requests you dont have access to.
They are autocreated with isSystem=true on update of status of the actual request.

### TimelineEntry Model

Adds a `TimelineEntry` model to track updates related to a `LendingRequest`.
Inherits from `BasisModel`.

**Fields:**
- `lending_request`: ForeignKey to the associated `LendingRequest`.
- `message`: Text message (user or system-generated).
- `is_system`: Boolean flag indicating if the entry was system-generated.
- `status`: When is_system, refers to the new  status of the request. Can not be manually set.

### TimelineEntryViewSet

Overrides the `create` method to return the full serialized `LendingRequest` (with updated timeline entries) instead of the newly created `TimelineEntry`.
Improves frontend context by returning the updated request state immediately after timeline updates.

#### `TimelineEntrySerializer`
Read-only serializer for displaying timeline entries.

**Fields:**
- `id`: Timeline entry ID
- `created_by`: Public user info of the creator
- `created_at`: Timestamp of entry creation
- `lending_request`: Associated `LendingRequest` ID
- `message`: Message content
- `status`: Optional status (from `LENDING_CHOICE_STATUSES`)
- `is_system`: Whether the entry is system-generated

#### `TimelineEntryCreateAndUpdateSerializer`
Write serializer for creating timeline entries with validation.

**Fields:**
- `id`: Entry ID (read-only)
- `lending_request`: Required FK to the request being updated
- `message`: Message text

**Validation:**
- Checks if the user has `VIEW` permission on the linked `LendingRequest` using object permissions.

## Tests:

- Leaving a message as the author of the request
- Leaving a message as a responsible user/group (handling the request)
- Not being permitted to leave a message as an unauthorized person (someone who does not fall into either of the above 2 categories)
- Updating the status of the request (either as a creator or handler) will autogenerate a system message and set updated_by to request.user

# Testing

Resolves ABA-1369


eikhr (changes requested):
Remember to write some tests!

jonasdeluna:
The tests are all written with 100% patch coverage.
They test:

- Leaving a message as the author of the request
- Leaving a message as a responsible user/group (handling the request)
- Not being permitted to leave a message as an unauthorized person (someone who does not fall into either of the above 2 categories)
- Updating the status of the request (either as a creator or handler) will autogenerate a system message and set updated_by to request.user

The person who integrates this in the front-end should (@itsisak):

1. Update the types and remove the old comment components
2. Create the action with actiontype LendingRequest.EDIT
3. (Add LendingRequest.EDIT to the redux slice on fetchActions
4. Make the UI for calling the action and displaying the entries
5. Done!

lego/apps/lending/serializers.py
```diff
@@ -59,6 +58,45 @@ def to_representation(self, value):
...
+    class Meta:
+        model = TimelineEntry
+        fields = (
+            "id",
+            "lending_request",
+            "message",
+        )
+
+    def validate(self, attrs):
+        attrs = super().validate(attrs)
```
Arashfa0301, line 88: Please make a type object for `attr` and apply it to this variable

lego/apps/lending/serializers.py
```diff
@@ -59,6 +58,45 @@ def to_representation(self, value):
...
+        lending_request = attrs.get("lending_request")
```
Arashfa0301, line 89: Please make a type object for `lending_request` and apply it to this variable

lego/apps/lending/serializers.py
```diff
@@ -59,6 +58,45 @@ def to_representation(self, value):
...
+        user = self.context["request"].user
```
Arashfa0301, line 90: Please make a type object for `user` and apply it to this variable

Arashfa0301 (approved):
Good job 🥰

eikhr (approved):
Looks good, but we will have to wait for the frontend to be ready before merging this as it has breaking changes, right?

jonasdeluna:
Yes, this will break the lending requests as of right now but if we just remove the commentdisplay/form and types frontend we can merge it without breaking. Everything is behind a featureflag and is not used yet :+1:

itsisak (approved):
Looks good!

PR #3814, opened 2025-03-28
Fix plausible statistics 

itsisak:
# Description

Seems like my previous "fix" did not work, so this fix should get statistics based on slug (if model has it) otherwise default to id. Also wrote some tests to ensure the functionality of the `request_plausible_statistics` function. 

# Testing


jonasdeluna (approved):
Nice!

eikhr (approved):

Arashfa0301 (approved):
nice

PR #3825, opened 2025-04-04
Send websocket message on new notification

eikhr:
# Description

Add a websocket message that the user has received a new notification. This allows the notification bell in the header to be updated instantly when the user receives a notification.

# Testing

I have tested that the websocket message is sent successfully with a couple of different notification types, and written some tests for the most common notifications.

Resolves ABA-1420


lego/apps/feeds/websockets.py
```diff
@@ -0,0 +1,25 @@
...
+
+    group = group_for_user(aggregated_activity.feed_id)
+    serializer = FeedActivitySocketSerializer(
+        {
+            "type": "SOCKET_NEW_NOTIFICATION",
+            "payload": aggregated_activity,
+        }
+    )
+    data = serializer.data
+    data["payload"] = FeedViewSet.attach_metadata([data["payload"]])[0]
```
jonasdeluna, line 24: Why is this done like this? Or is there no other way

jonasdeluna (approved):
Nice work.
I would love to read through all the websocket code more properly and try to understand it. Nice with the generic serializer maybe we could use websocket more though I guess events is the main usecase and its done already.

lego/apps/feeds/websockets.py
eikhr: It's a hack. `FeedViewSet` runs `attach_metadata` on GET requests to feeds in order to add data about the relevant entities (like the Meeting for a MeetingInvitation), but in the ViewSet it is run on the list of fetched feed-tings.
In the notification we have an object with a field "payload" that is one such feed-thing, so in order to make the existing `attach_metadata` method work for adding the required metadata I do the clumsy property access and the list stuff in this line.

Viljen789 (approved):
Sigma!

PR #3826, opened 2025-04-05
Automatic email to users on lending request status change

jonasdeluna:
# Description

Adds automatic email on status change for lending requests:

* Admins get notified when the user cancels
* The user gets notified when the admin updates the status

# Testing

Resolves ABA-1397


ch0rizo:
Open question: Shouldn't the admins only get email notification, if the user cancel an already approved request?

I'm just worried that this may cause a lot of spam and unnecessary many notifications for the admins.

Arashfa0301 (approved):
Good job

eikhr:
I haven't thourougly review all the code, but it looks very good!

I also think they should get a front-end notification in the bell thing in the header for this. It should be easy to implement, just have a look at how it is done for announcements https://github.com/webkom/lego/blob/master/lego/apps/notifications/action_handlers.py . We usually do both feed-notification and email/push when we send out notifications:)

jonasdeluna:
> Open question: Shouldn't the admins only get email notification, if the user cancel an already approved request?
> 
> I'm just worried that this may cause a lot of spam and unnecessary many notifications for the admins.

With the current implementation the admins only get a notification on cancel.
The user may get more notifications about updates though.

jonasdeluna:
> I haven't thourougly review all the code, but it looks very good!
> 
> I also think they should get a front-end notification in the bell thing in the header for this. It should be easy to implement, just have a look at how it is done for announcements [`master`/lego/apps/notifications/action_handlers.py](https://github.com/webkom/lego/blob/master/lego/apps/notifications/action_handlers.py) . We usually do both feed-notification and email/push when we send out notifications:)

I might look at that later but I tried adding push notifications for achievements but couldnt figure it out for a long time. Maybe we could look at it at a meeting but in the meantime I think this is the last "strictly necessary" feature for the lending system to be used.

ch0rizo (approved):

PR #3833, opened 2025-04-17
Remove tag validation because theres no need for that

jonasdeluna:
# Description

Editing event tags from the editor for achievements will fail without this.
No need for this and trophies use the tags trophy:gala and trophy:genfors so they need to be allowed to set from serializer

This is needed for https://github.com/webkom/lego-webapp/pull/5556 to work properly.

# Testing

Resolves ABA-1436


Viljen789 (approved):

Viljen789:
Nice!

norbye (commented):
Is there any reason why you're not keeping the validator but allowing the use of colons?

jonasdeluna:
> Is there any reason why you're not keeping the validator but allowing the use of colons?

That might be a better option I'll change it to that :+1:

ch0rizo (approved):

PR #3834, opened 2025-04-30
Add HS-rep group to LDAP sync groups

juniwbjerde:
# Description
Created a new group for HS-rep in shell, and added it to the `LDAP_GROUPS` (the hardcoded list of groups that is synced to ldap, aka wiki), so that the HS-rep can have access to wiki without being showed as a HS-group member and getting extra permissions etc.

Resolves ABA-1439


ch0rizo (commented):
HUGE! 🪨

ch0rizo (approved):

ivarnakken (approved):

PR #3835, opened 2025-05-12
Send websocket message for comments

itsisak:
# Description

Send websocket messages for new comments and deleted comments. 

Currently a bit unsure of how i should implement the groups.

There should imo be a seperate group for each comment section, such that you only get messages from those you have access to (eg. "comments-events.event-1"). The issue is that always connecting to all groups a user has access to seems inefficient since there exists so many events, meetings etc. 

# Testing

Resolves ABA-1465


lego/utils/serializers.py
```diff
@@ -14,11 +14,13 @@ class GenericRelationField(serializers.CharField):
...
     def __init__(self, *args, **kwargs):
-        kwargs["write_only"] = True
         super().__init__(*args, **kwargs)
 
     def to_representation(self, value):
-        return None
+        try:
+            return instance_to_string(value)
+        except Exception:
+            pass
```
itsisak, line 23: Dont know why `GenericRelationField` was write-only, it prevented comments from returning `content_target`. If it should stay how it was, it is easy enough to add it to the websocket meta data instead of this.

PR #3849, opened 2025-10-04
Feat: Command palette

ch0rizo:
# Description

Backend PR for command palette. Palette tracks how relevant a command is to show suggested commands to use.

# Testing

Tested with insomnia and both POST and GET requests works fine.

Resolves ABA-1555


lego/apps/users/validators.py
```diff
@@ -3,6 +3,38 @@
...
+    "trophies",
+    "theFund",
+    "developerBlog",
+    "settings",
+    "createMeetingNotice",
+    "createReceipt",
+    "createQuote",
+    "createAlbum",
+    "logout",
+}
```
itsisak, line 27: Thought we did not want this?

ch0rizo: If not there it will mess with the redux state. But there is some frontend "validation", to keep it from displaying. Just though it would be best practice🤷‍♂️

itsisak (approved):
This is getting very clean!

Very very minor question. Is there a particular reason that the serializer returns the whole array off suggestions (including the buffer) while the endpoint in `UserCommandViewSet` only returns the 3 suggestions?

If you always only need the results defined in your `user.get_command_suggestions()` method I suggest using this function in the serializer as well. It should be very easy to implement just see how eg. penalites uses a custom method in the same serializer.

With or without the above fix I think this is ready to merge!

ch0rizo:
> This is getting very clean!
> 
> Very very minor question. Is there a particular reason that the serializer returns the whole array off suggestions (including the buffer) while the endpoint in `UserCommandViewSet` only returns the 3 suggestions?
> 
> If you always only need the results defined in your `user.get_command_suggestions()` method I suggest using this function in the serializer as well. It should be very easy to implement just see how eg. penalites uses a custom method in the same serializer.
> 
> With or without the above fix I think this is ready to merge!

Thank you, I was wondering why endpoint gave 3 suggestions, while redux had 5😅🎉

PR #3855, opened 2025-10-11
Add: Lending category

ch0rizo:
# Description

Added categories for lendingObjects, so that we call filter on categories👍 

**I am unsure which categories that are relevant... Please come with some suggestions**😄 

# Testing

Resolves ABA-1557


falbru (approved):
Looks good! We probably need to change the categories later as Isak said

PR #3896, opened 2026-02-07
Parsing redirect uri wildcard

### LINKED ISSUE #3895: Wildcard and space separated for redirect URIs
Add support for wildcard and space separated values in Application Redirect URIs

Viljen789:
# Description
Added wildcards and support for both space and comma separated links in application redirect uris, and supporting tests.

> [!WARNING]
readme requested this feature, and even though i have some simple checks to avoid abuse of wildcards, this might become dangerous if combined with another request from readme which was that everyone can add applications

# Testing

Added tests to check for new scenarios with wildcard, to make sure it works as expected and it catches the worst mistakes. 

Resolves #3895


ch0rizo (commented):
Co Pilot do not like this PR💀

PR #3913, opened 2026-02-26
Delete cascade lending

### LINKED ISSUE #3887: Delete lendable object
Backend functionality is implemented. Easy to implement button to delete

Problems: 
- When an item is deleted it list /lending is not updated, showing deleted items. Same with "Innboks".

Solution:
- Look how events does it?
- Refetch list to remove stale items. But refetch on every reload is not optimal.

ch0rizo:
# Description

Deleting a lendable object did not cascade to lending request and timeline. This PR fixes this.

# Testing

Resolves #3887

### DIFF
--- a/lego/apps/lending/tests/test_lendableobject_api.py
+++ b/lego/apps/lending/tests/test_lendableobject_api.py
@@ -328,15 +328,23 @@ class DeleteLendableObjectTestCase(BaseAPITestCase):
             current_user=self.user,
         )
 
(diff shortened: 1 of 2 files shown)


jonasdeluna (approved):
Nice! On delete setting did not work I assume and the custom code was needed?

ch0rizo:
> Nice! On delete setting did not work I assume and the custom code was needed?

Yes. So because of soft delete since no real delete happens, Django’s deletion collector never runs, so CASCADE does not fire

PR #4114, opened 2026-04-13
Add lending archiving

Bestem0r:
# Description

Adds archiving of lending requests to clear up inbox.

[Frontend PR](https://github.com/webkom/lego-webapp/pull/5937).

# Result

- [x] Changes look good on both light and dark theme.
- [x] Changes look good with different viewports (mobile, tablet, etc.).
- [x] Changes look good with slower Internet connections.

https://github.com/user-attachments/assets/e8ab0b85-238a-4ec5-8d51-7d3714d3ee46

# Testing

- [x] I have thoroughly tested my changes.****


ch0rizo (changes requested):
Anyone who can PATCH can archive any request. The archived field is added directly to `LendingRequestCreateAndUpdateSerializer` without any permission guard. The existing validate() method only checks permissions on status changes - it doesn't restrict who can set archived. This means:

  - An admin/responsible person could archive a requester's request.
  - The requester could archive someone else's request if they somehow get access

lego/apps/lending/serializers.py
ch0rizo, line 206:
```suggestion
  if "archived" in attrs and self.instance:
      if self.instance.created_by != user:
          raise serializers.ValidationError(
              {"archived": "You can only archive your own requests."}
```

ch0rizo (changes requested):

ch0rizo (approved):
Great!

PR #4243, opened 2026-08-18
Replace poetry and tox with uv

### LINKED ISSUE #4203: Migrate to uv package manager
Image

Do I need to say more?

ch0rizo:
# Description

`uv` replaces both poetry and tox. `pyproject.toml` moves to PEP 621, `uv.lock` replaces `poetry.lock`, and Drone steps call `uv run` directly.

**The lock is a faithful port, not a re-resolve.** A fresh `uv lock` pulls everything to latest and produced **88 differing transitive versions** (`urllib3 1.26.19 → 2.0.7`, `protobuf 4.22.3 → 7.35.1`, `autobahn 23.1.2 → 26.7.1` in the channels stack, numpy, matplotlib, markupsafe). That is years of accumulated drift and does not belong in a tooling change, so the resolution was seeded from `poetry.lock` and verified package by package: **197 packages, 0 version mismatches, nothing added or dropped.** Only tox and its exclusive dependencies leave. Those 88 upgrades are worth taking separately.

| | poetry | uv |
|---|---|---|
| lock from scratch, same deps | 46.4s | **2.3s** |
| create a 197-package venv, warm cache | - | **0.6s** |
| CI tooling per step | poetry + tox | uv |

The base image stays `abakus/lego-testbase:python3.11`; each step installs uv with pip. Switching to `ghcr.io/astral-sh/uv:python3.11-bookworm` and dropping that install is a worthwhile follow-up, kept out of here so the image is not a second moving part.

# Testing

Every Drone step run locally as written, against postgres 17 and redis:

- tests: exit 0, **1267 tests, OK (skipped=10)**
- coverage: **84%**, 13 parallel data files combined, 75% gate passes, which is what proves the `COVERAGE_PROCESS_START` + `PYTHONPATH` wiring survived
- flake8 0 findings in 1.1s, isort clean, black clean, missing-migrations clean, docs build OK
- Docker image builds and boots: `uwsgi`, `celery`, `daphne` all resolve to `/usr/local/bin`, versions match the lock, weasyprint imports. Rebuilt after the `.dockerignore` fix to confirm `/app/.venv` is gone.

Resolves #4203


jonasdeluna (approved):
👏📈📈

PR #4293, opened 2026-08-30
Fix pool counter drift on bump paths

### LINKED ISSUE #4287: Event pool counters disagree with the actual registration count
Example: `PoolCounterNotEqualToRegistrationCount: Pool 2714` for event 4151 was supposed to have 29 registrations, but has 30. Event Ping med Bekk, pool 2. og 3. klasse.

The denormalised pool.counter is what capacity, waiting-list bumps and admission decisions are computed from. When it drifts low, a pool over-admits; when it drifts high, someone on the waiting list never gets bumped.

ch0rizo:
## Problem

`Pool.counter` is what `add_to_pool` admits on, but three of the six paths that move a registration into a pool never updated it: `early_bump`, `bump_on_pool_creation_or_expansion` and `rebalance_pool`.
Counters drifted below the real registration count, so full pools kept admitting past capacity and the nightly checker failed.
`increment()`/`decrement()` also did a read-modify-write on the in-memory instance, so a caller holding a stale `Pool` wrote back a wrong absolute value.

## Fix

- `Registration.move_to_pool()` is now the single place that changes a registration's pool, adjusting both counters in one transaction. `bump`, `early_bump`, `bump_on_pool_creation_or_expansion` and `rebalance_pool` go through it, so no call site has to remember the pairing.
- `Pool.increment()`/`decrement()` use an `F()` expression, correct regardless of staleness or concurrency.
- Migration `0048` recomputes every counter from the registrations that exist, repairing rows written before this fix.
- The checker collects all mismatches instead of raising on the first, and logs `pool_id`, `event_id`, `counter` and `registration_count` as structured fields. It previously logged only the pool name.

Follow-ups: merged single-pool events still skip the decrement on unregister, so seats sit empty; `rebalance_pool` reports `bumped` when nobody was bumped; `check_for_bump_on_pool_creation_or_expansion` takes no lock; and longer term `Pool.counter` should go away entirely.

# Testing

Resolves #4287

### DIFF
--- a/lego/apps/events/exceptions.py
+++ b/lego/apps/events/exceptions.py
@@ -78,11 +78,37 @@ class EventNotReady(ValueError):
 
 
 class PoolCounterNotEqualToRegistrationCount(ValueError):
-    def __init__(self, pool, registration_count, event):
+    MAX_POOLS_IN_MESSAGE = 10
+
+    def __init__(self, mismatches):
+        """
+        :param mismatches: list of (pool_id, event_id, counter, registration_count)
+        """
+        # Celery and pickle rebuild exceptions as cls(*args), and args hold the message.
+        if isinstance(mismatches, str):
+            self.mismatches = []
+            super().__init__(mismatches)
+            return
+
+        self.mismatches = mismatches
+        details = [
+            f"Pool {pool_id} for event {event_id} was supposed to have "
+            f"{counter} registrations, but has {registration_count}"
+            for pool_id, event_id, counter, registration_count in mismatches
+        ]
+
+        if len(details) == 1:
+            super().__init__(f"{details[0]}!")
+            return
+
+        listed = details[: self.MAX_POOLS_IN_MESSAGE]
+        remaining = len(details) - len(listed)
         message = (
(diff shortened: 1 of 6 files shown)


ch0rizo:
We should also have more focus on writing better exception handlers and implement them more frequent specifically for `Events` - its arguably the most critical business logic on abakus.no. Its also very useful having Sentry logs for Events. The thing that makes error actually actionable. Cuts a lot of debugging time.

Generally `Events` are so much work to make some few changes. It's messy and theres multiple places where the same calculations happens independently, which makes it frustrating and hard to implement a concrete fix.

I am really down to take a look at this deeper, and basically rewriting this to be more maintainable, readable and robust!

jonasdeluna (approved):
Nice! But what was the rationale behind making the resync a migration and not running it once in shell?

ch0rizo:
> Nice! But what was the rationale behind making the resync a migration and not running it once in shell?

I think its nice to have this reconciliation script declared or logged in some kind of way. Might be overkill🤷‍♂️

itsisak (approved):
Can we merge this to prevent wrong registrations

PR #4303, opened 2026-09-07
Properly leave ws groups

itsisak:
Makes it possible for client to know which group it left to keep matching state and avoid errors


jonasdeluna (approved):

ch0rizo (approved):
What does this do? Just appends the groupID?

itsisak:
> What does this do? Just appends the groupID?

Yea, so that the client knows which group this message refers to and can update status, previously group status was just left at pending since updating status for a given group fails without that groupID.

PR #4314, opened 2026-09-11
Remove the pool.counter

### LINKED ISSUE #4313: Fix pool counter drift

jonasdeluna:
# Description

Remove the pool.counter, replace with COUNT query.
This was tested locally on 3k regs being about 80ms slower. It should be performance wise fine and avoids all drift.
Also properly set registration failure when retry fails.
Centralize a lot of code regarding bumping/select_for_bump into one method.
Ensure always using transaction.atomic on updating the pools and regs.

# Testing

Some load tests and updated old tests.

Resolves #4313

### DIFF
--- a/lego/apps/events/tasks.py
+++ b/lego/apps/events/tasks.py
@@ -2,7 +2,6 @@ import uuid
 from datetime import timedelta
 
 from django.db import IntegrityError, transaction
-from django.db.models import Q
 from django.utils import timezone
 
 import stripe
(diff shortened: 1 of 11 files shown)


lego/apps/events/tasks.py
```diff
@@ -142,6 +136,7 @@ def async_register(self, registration_id, logger_context=None):
             exception=e,
             registration_id=self.registration.id,
         )
+        raise
```
itsisak: ?

ch0rizo (approved):
This is great! Moving in the right direction. If I understand we loose the registration count ("cache"), that is what causes the higher latency?

itsisak (approved):
Nice to have some tests to ensure concurrent behavior stays correct. We should also test that registrations work fine with this in staging too.

jonasdeluna:
> This is great! Moving in the right direction. If I understand we loose the registration count ("cache"), that is what causes the higher latency?

Yes, reading that value was quicker but hopefully there should be not drift issues at all now and the delay seems quite minor. We could make everyone in Webkom do a test register to a small event just to make sure in staging perhaps.

jonasdeluna:
Should we test it out in staging? @itsisak

itsisak:
> > This is great! Moving in the right direction. If I understand we loose the registration count ("cache"), that is what causes the higher latency?
> 
> Yes, reading that value was quicker but hopefully there should be not drift issues at all now and the delay seems quite minor. We could make everyone in Webkom do a test register to a small event just to make sure in staging perhaps.

If the slowdown this introduces is to big I think we should revisit ways to shorten it. In any ways I also think we should update the user feedback for when the request is being handled.

PR #4327, opened 2026-09-16
Chore: Fix context field and serializer for feed-notifications

### LINKED ISSUE #4311: Feed notifications responses do not match with the generated schema
Feed notifications endpoints have responses that do not match with the generated schema from OpenAPI.

**/api/v1/feed-notifications/:**
Context field shows up in the response body, but not in the schema. This means trying to retrieve the context after fetching does not work.

Image

Image

**/api/v1/feed-notifications/notification_data/:**
The response body shows unreadCount and unseenCount, but the schema shows AggregatedMarkedFeed. After calling the notifications_data endpoint, you get the AggregatedMarkedFeed fields not NotificationData fields.

Image

Image

TonyWorep:
# Description

The context field was previously attached to feed responses after serialization via `FeedViewSet.attach_metadata()`, so it never appeared in the generated schema despite always being present in the response body. It has now been moved.

Notification_data had no serializer_class set so the viewset defaulted to the AggregatedMarkedFeedSerializer. We added a NotificationsDataSerializer and set it via serializer_class on the action.

# Testing

Wrote FeedContextTestCase to cover the context field.

Resolves #4311

### DIFF
--- a/lego/apps/feeds/context.py
+++ b/lego/apps/feeds/context.py
@@ -1,0 +1,17 @@
+def collect_refs(obj):
+    """
+    Given one aggregated feed object, return the set of actor/object/target
+    strings referenced by its activities.
+    """
+    refs = set()
+    for activity in obj.activities:
+        for attr in ("actor", "object", "target"):
+            ref = getattr(activity, attr, None)
+            if ref:
+                refs.add(ref)
+    return refs
+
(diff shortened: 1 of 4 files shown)


jonasdeluna (approved):
Nice!

PR #4334, opened 2026-09-18
Add lendable bundle

### LINKED ISSUE #4333: Add LendableBundle
Make lendableojbects contain themself

jonasdeluna:
# Description

Makes lendableoebjects able to be nested like a tree, allowing for bundles.
They can be lended similarly to lendableobjects and will check availability using union of the children.
Simplify some code and move availability to the model instead of the view.

Also fixes the recursive delete bug.

Does not support having either one of two items fulfilling the same item in a bundle and requesting either one of them.

# Testing

Resolves #4333


lego/apps/lending/models.py
```diff
@@ -29,10 +32,22 @@ class LendableObject(BasisModel, ObjectPermissionsModel):
...
         permission_handler = LendableObjectPermissionHandler()
 
+    def delete(self, using=None, force=False) -> tuple[int, dict[str, int]]:
+        if force:
+            return super().delete(using=using, force=force)
+
+        with transaction.atomic():
+            for lending_request in self.lendingrequest_set.all():
+                lending_request.delete(force=False)
+            return super().delete(using=using, force=force)
```
falbru, line 49: I don't necessarily think that if you were to delete a bundle, it should delete all items of that bundle. What if you still want the items to be lended, but don't want to offer the bundle anymore? What if there are multiple bundles that own the same object, wouldn't this just delete the item anyway?

lego/apps/lending/models.py
```diff
@@ -42,14 +57,106 @@ def can_lend(self, user: User) -> bool:
...
+        for lending_request in overlapping_requests:
+            range_start = max(lending_request.start_date, start_of_month)
+            range_end = min(lending_request.end_date, start_of_next_month)
+            created_by = lending_request.created_by
+            fullname = created_by.get_full_name() if created_by else None
+            username = created_by.username if created_by else None
+            ranges.append(
+                (range_start.isoformat(), range_end.isoformat(), fullname, username)
+            )
+        return ranges
```
falbru, line 123: I think this should return a class and not a list of tuples, since tuples are arbitrary and unstructured. What if it just returns the overlapping_requests? I don't think the range_start and range_end modifications are meaningful, and then it's up to the serializer how to display the data.

falbru (commented):
Very nice refactor! You say it will behave like a tree, but doesn't the many2many relation allow it to be cycles and allow one lendable object to be apart of multiple bundles? If so, this should be taken into account for the availability and delete methods.

